PR #67, opened 2024-03-26
Create Database interface - Firestore/Mock implementations

Plouc314:
Create a `Database` interface with basic read/write/delete functions.
Create Firestore and mock implementations.

Close #52

PR #68, opened 2024-03-26
Data model availability

### LINKED ISSUE #54: Data model: Availability
Define how the availabilities of a crew/pilot are represented and stored.

**Frontend**
* Define what attributes are needed
* Think of how you will query them, based on what criteria. 

**Database**
* Think of how to represent the data to have them easily query-able
* Think of potential similarities with flight temporal informations

Plouc314:
Based on #67 for database setup.

Split data model into "schema" and "model", one for the database representation and one for the frontend representation.

Create schema, model and table classes for the Availability data model.

Close #54

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/database/MockDatabase.kt
+++ b/app/src/main/java/ch/epfl/skysync/database/MockDatabase.kt
@@ -8,6 +8,10 @@ class MockDatabase : Database {
   private var idCounter = 0
   private var state = hashMapOf<String, Any>()
 
+    fun getState(): HashMap<String, Any> {
+        return state
+    }
+
   private fun getKey(path: String, id: String): String {
     return "$path/$id"
   }
@@ -21,7 +25,7 @@ class MockDatabase : Database {
     val key = getKey(path, idCounter.toString())
     state[key] = item
     idCounter += 1
-    onCompletion
+    onCompletion()
   }
 
   override fun <T : Any> get(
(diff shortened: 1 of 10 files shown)


RAFInery:
a few comments:
- They actually organize their availabilities in two time **slots** per day (morning/evening). Since the exact working hours are only published 12h in advance and may vary a lot, staff must give their availability for the entire morning and/or evening of a day.
- there should be 3 types of availabilities for a slot: 
   - "available", 
   - "maybe" (admin will double check with crew member before assigning)
   - not available

- I also suggest that we reference (with respect to the chosen structure of the  time slots e.g.  daily, weekly, ...) the collection of availabilities directly from the User. I.e. every user would have a property "availability". 
For the admins view, we'll collect those availabilities into one collection so that he can query them.

Here as a reference again the current structure of their excel sheet
Screenshot 2024-03-26 at 13 36 55

RAFInery:
nice. Jep, regarding the personid, we might as well remove it from the attributes of **Availability** (to simplify the class) if we reference it from the users and add it only in the request sent to firebase

Plouc314:
I see what you mean, for that we first need to do the person data model but I totally agree

RAFInery:
Generally many staff members will give not contiguous availabilities (since they work part-time). However, I like the idea of giving a range of availabilities (especially for full-time workers). Let's assume a user wants to be available every sunday or a whole week, every day except mondays...there should be an easy way to do so instead of setting the availabilities individually. I'll create an issue for that

RAFInery:
exactly but we might first want to discuss/brainstorm what kind of recurrences we want to allow, how to undo them etc.

RAFInery (approved):

PR #71, opened 2024-03-27
Navigation and scaffold

### LINKED ISSUE #62: UI: Scaffold and navigations
* Create the basic scaffold used for the home page screens, try to make it simple and generic, we ideally want to reuse it as much as possible.
* Setup the navigation logic between screens, also take into account the login screen.

Baribal02:
The nested navigation graph implemented in MainActivity has a sub-graph homeGraph that englobes the screens used with the bottom app bar.

I regrouped the different parts of the implementation of the NavigationBar in a single file

I changed the screens used in the homeGraph to implement the Scaffold and navigate between them using the NavigationBar.

Close #62


app/src/main/java/ch/epfl/skysync/screens/LoginScreen.kt
```diff
@@ -44,7 +53,10 @@ fun LoginScreen(signInLauncher: ActivityResultLauncher<Intent>, user: FirebaseUs
 
           if (user != null) {
             // logged in
-            Text(text = "Welcome")
+            navController.navigate(Route.HOME) {
```
Plouc314, line 56: Is there a reason why the `navigate` function is called here instead of in the `MainActivity.onSignInResult` callback ?

app/src/main/java/ch/epfl/skysync/navigation/Route.kt
```diff
@@ -0,0 +1,11 @@
+package ch.epfl.skysync.navigation
+
+object Route {
+  const val MAIN = "Main screen"
```
Plouc314, line 4: I'm not sure I understand what the `Main screen` represent

app/src/main/java/ch/epfl/skysync/screens/Chat.kt
```diff
@@ -1,32 +1,34 @@
...
-      contentAlignment = Alignment.Center) {
-        Text(
-            text = "Chat",
-            fontSize = MaterialTheme.typography.h3.fontSize,
-            fontWeight = FontWeight.Bold,
-            color = Color.White)
-      }
+fun ChatScreen(navController: NavHostController) {
+  Scaffold(modifier = Modifier.fillMaxSize(), bottomBar = { BottomBar(navController) }) { padding ->
+    androidx.compose.material3.Text(
```
Prolas, line 20: It would be better to import it at the beginning and replace the "import androidx.compose.material.Text" with "import androidx.compose.material3.Text"

Baribal02: Yes, I fixed it it was due to the import of material instead of material 3

app/src/main/java/ch/epfl/skysync/navigation/Route.kt
Baribal02: "Main screen" is maybe not quit a good name... It is in fact the graph route that contains the four screens : Home, Flight, Chat and Calendar.  As it is a nested graph to access to these four screens from outside (in our case from the login) we need to navigate to this MAIN node which will automatically redirect to the Home screen

app/src/main/java/ch/epfl/skysync/screens/LoginScreen.kt
Baribal02: My problem is how to deal with someone who is already connected. With the actual implementation it does not seem possible but it would be suitable that we do not have to connect every time we access the app. So I haved two choices either making it this way or put a condition on the start destination of the navigation graph. What do you think is better ?

app/src/main/java/ch/epfl/skysync/navigation/Route.kt
Plouc314: Okay, that makes sense, then "Main graph" would be clearer to me

app/src/main/java/ch/epfl/skysync/MainActivity.kt
```diff
@@ -36,6 +41,16 @@ class MainActivity : ComponentActivity() {
           this.onSignInResult(res)
         }
 
-    setContent { SkySyncTheme { LoginScreen(signInLauncher = signInLauncher, user = user.value) } }
+    setContent {
+      SkySyncTheme {
+        val navController = rememberNavController()
+        NavHost(navController = navController, startDestination = Route.LOGIN) {
```
mathieumognol, line 47: Just to be sure will it always take us to login screen when launching app or if we are already logged in we will  end up at home screen directly?

Baribal02: I've modified this after this commit to take us to home screen if we already are logged in

Zep3k (approved):
The code is good, it looks clean and well organized, but some documentation would make reading it easier

app/src/main/java/ch/epfl/skysync/navigation/MainGraph.kt
```diff
@@ -0,0 +1,25 @@
...
+fun MainGraph(
+    navHostController: NavHostController,
+    signInLauncher: ActivityResultLauncher<Intent>,
+    user: FirebaseUser?
+) {
+  NavHost(
+      navController = navHostController,
+      startDestination = if (user == null) Route.LOGIN else Route.MAIN) {
+        homeGraph(navHostController)
+        composable(Route.LOGIN) { LoginScreen(signInLauncher = signInLauncher) }
```
Prolas, line 23: signInlauncher comes from MainActivity.  Wouldn't it be better to create a LoginScreenModel that contains the sign-in logic ?

app/src/androidTest/java/ch/epfl/skysync/HomeNavigationTest.kt
```diff
@@ -0,0 +1,62 @@
...
+
+  @Test
+  fun verifyHomeIsStartDestination() {
+    val route = navController.currentBackStackEntry?.destination?.route
+    Assert.assertEquals(route, Route.HOME)
+  }
+
+  @Test
+  fun routeIsRightIfClickOnCalendar() {
+    composeTestRule.onNodeWithText("Calendar").performClick()
```
Prolas, line 41: I did not know this onNodeWithText function. This simplifies testing, but the text must be unique throughout the application to work properly.

Prolas:
I updated the test documentation accordingly to your tests

app/src/main/java/ch/epfl/skysync/navigation/Route.kt
```diff
@@ -0,0 +1,12 @@
+package ch.epfl.skysync.navigation
+
+/** Define the paths used for the navigation in the app */
+object Route {
+  const val MAIN = "Main graph"
```
mathieumognol, line 5: What's the difference between "Main" Route and "Home" Route?

Plouc314: See [here](https://github.com/SkySync-EPFL/skysync/pull/71#discussion_r1541202006) (I had the same question)

app/src/main/java/ch/epfl/skysync/navigation/MainGraph.kt
Plouc314: We'll see about that later

PR #73, opened 2024-03-27
Set up environment for testing

### LINKED ISSUE #63: Explore testing
* See how to create/run logic tests (which are not directly linked to some UI elements)
* See how to create/run UI tests

The testing phase of all other issues is dependent on this task, so we expect to not have everything properly tested for this sprint.

Prolas:
Close #63 

Add all dependencies to create tests and generate a jacoco test report.
Add a documentation on how to create tests.

I'm uncertain about the specific file structure required for organizing the UI tests. For each class we want to test, we need to create at least two corresponding classes. One class is placed in the same location as its counterpart, located in app/src/main/java/ch/epfl/skysync. As for the other class, should we include it in a tests folder within app/src/androidTest/java/ch/epfl/skysync?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -12,3 +12,142 @@ An application that simplifies and optimizes staff interactions and planning in 
 | |
 | - |
 | [Standup 1](https://docs.google.com/spreadsheets/d/1m8U5FLoCYPnLGZMNuJ6yGFX6ynVkkjzNFThc0HQJ9-g/edit#gid=0) |
+
+## Testing
+
+### UI Testing: Global Description
+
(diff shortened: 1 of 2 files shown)


hacknstack (approved):
The comments provide clear instructions on how to set up UI tests using Compose and how to create backend tests. This is helpful for us because we're developers who are sort of new to testing in Android or using Compose.
The code is well-organized with clear sections for UI testing and backend testing. Each section explains the steps to set up testing in a structured manner.
The could be further improved with more examples or explanations for certain concepts in the future, depending on the specific case that we might encounter often during the proje

PR #75, opened 2024-03-28
Flight/User/Calendar model

### LINKED ISSUE #53: Data model: Person
Define how a person is represented and stored.

**Frontend**
* What are the difference between a crew member and a pilot, in term of attributes that need to be stored.
* How do we want to represent admins, do we want to stored them.
* What are the similarities between the different types of person (crew/pilot/admin)
* Think of a hierarchical data model to represent the data on the app logic side (with parent/child classes)

**Database**
* Think of how to represent the data on the database (in #72)

### LINKED ISSUE #55: Data model: Flight
Define how a flight is represented and stored.

* Think of the information needed for a flight to take place
* Think of the relations between a flight and other data models (for example: a flight might have one or more crews assigned): are they mandatory, one-to-one, one-to-many
* Think of if it makes sense to break the flight model into multiple smaller logical components

The specifics of the database will be discussed later.

### LINKED ISSUE #74: Data Model: Calendar
part of a user is its availabilities and assigned flights

- think how to structure the calendar
- consider abstract implementations to reuse for both availabilities and flights
- add basic functions to access and manipulate the calendar

RAFInery:
modelisation of the three main data models **Flight**, **User**, **Calendar**

closes #55, 
closes #53 
closes #74

(the issues were pretty interleaved so I treated them together)

# Flight: 
- consists of a main interface and different implementations depending on current stage of flight (planned, confirmed, done). This allows to reduce the complexity and weight of flight objects (e.g. a planned flight does not need a property for flight duration) 
- there is a first implementation PlannedFlight
- many properties of a flight were implemented with data classes and enumerates to control the values to choose from
# User: 
- the User interface defines the main properties of a user
- so far we distinguish 3 types of users: Crew, Pilot, Admin that will be able to see different screens and can take different roles in a flight (e.g. a Crew member should not be assigned as a pilot to a flight)

# Calendar: 
- an abstract modelisation of the calendar was given that can be reused to store assigned flights (1 flight per slot), scheduled flights (multiple flights per slot) and availabilities
- a more specific definition was implemented for both flights and availabilities

A general overview of the data model structure:


app/src/main/java/ch/epfl/skysync/dataModels/calendarModels/CalendarModel.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import java.time.LocalDate
+import java.time.temporal.ChronoUnit
+
+abstract class CalendarModel<T: CalendarViewable>{
+    protected val cells: MutableList<T> = mutableListOf()
+    var size: Int = cells.size
+        private set
+
+    abstract fun initForRange(from: LocalDate,
```
Plouc314, line 11: Documentation is needed (it is even more important for abstract functions).

app/src/main/java/ch/epfl/skysync/dataModels/calendarModels/CalendarModel.kt
```diff
@@ -0,0 +1,78 @@
...
+        for (i in 0.. numberOfDays) {
+            for (timeSlot in TimeSlot.entries) {
+                cells.add(constructor(currentDate, timeSlot))
+                update_size()
+            }
+            currentDate = currentDate.plusDays(1)
+        }
+
+    }
+    private fun update_size(){
```
Plouc314, line 32: Instead of creating an attribute `size` and update it each time `cells` is updated, why not have a method `getSize()` (or just `size()`):
```kotlin
fun size(): Int {
    return cells.size
}
```

app/src/main/java/ch/epfl/skysync/dataModels/calendarModels/CalendarModel.kt
```diff
@@ -0,0 +1,78 @@
...
+        private set
+
+    abstract fun initForRange(from: LocalDate,
+                              to: LocalDate,)
+
+    /**
+     * @param from inclusive
+     * @param to inclusive
+     */
+    protected fun initForRangeSuper(from: LocalDate,
```
Plouc314, line 18: Documentation is needed

app/src/main/java/ch/epfl/skysync/dataModels/calendarModels/CalendarModel.kt
```diff
@@ -0,0 +1,78 @@
...
+            }
+                )
+            {
+                throw IllegalArgumentException("Cannot add cells for the same date and time slot twice")
+            }
+            cells.add(t)
+            update_size()
+        }
+    }
+    fun setByDate(date: LocalDate,
```
Plouc314, line 49: This should really be called `updateByDate` as it only update an existing cell and never create a new one

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/Balloon.kt
```diff
@@ -0,0 +1,6 @@
+package ch.epfl.skysync.dataModels.flightModels
+
+data class Balloon(val name:String, val qualification: BalloonQualification)
```
Plouc314, line 3: This should have an `id` field (even if never used by the front-end, it is needed for the database, especially in a case like `flight.setBalloon(balloon)`)

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/Basket.kt
```diff
@@ -0,0 +1,7 @@
+package ch.epfl.skysync.dataModels.flightModels
+
+data class Basket(val name:String, val hasDoor: Boolean)
```
Plouc314, line 3: This should have an id field (even if never used by the front-end, it is needed for the database, especially in a case like ```flight.setBasket(basket)```)

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/Team.kt
```diff
@@ -0,0 +1,20 @@
+package ch.epfl.skysync.dataModels.flightModels
+
+class Team {
```
Plouc314, line 3: Is there a reason why `Team` is not a data class ?

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/Vehicle.kt
```diff
@@ -0,0 +1,5 @@
+package ch.epfl.skysync.dataModels.flightModels
+
+data class Vehicle(val name:String)
```
Plouc314, line 3: This should have an `id` field (same reason as above)

app/src/main/java/ch/epfl/skysync/dataModels/userModels/User.kt
```diff
@@ -0,0 +1,19 @@
...
+interface  User {
+    val firstname: String
+    val lastname: String
+    val userId: String
+    val availabilities: AvailabilityCalendar
+    val assignedFlights: FlightCalendar
+
+    fun addRoleType(roleType: RoleType): User
+
+    fun hasRoleType(roleType: RoleType): Boolean
```
Plouc314, line 16: Why not have `roleTypes` defined in `User`? That way `hasRoleType`, `addRoleType` could be define in `User` too.

app/src/main/java/ch/epfl/skysync/dataModels/userModels/User.kt
```diff
@@ -0,0 +1,19 @@
+package ch.epfl.skysync.dataModels.userModels
+
+import ch.epfl.skysync.dataModels.calendarModels.AvailabilityCalendar
+import ch.epfl.skysync.dataModels.calendarModels.FlightCalendar
+import ch.epfl.skysync.dataModels.flightModels.RoleType
+
+interface  User {
+    val firstname: String
+    val lastname: String
+    val userId: String
```
Plouc314, line 10: This could be `id` to be consistent with the other models

Prolas:
Do the user have an attribute role which is either crew, pilot or admin ? If yes, what about more specific roles (maitre fondu) ? Should we do 2 types of role ? One per user (crew, pilot or admin) to determine the permission (user information, ...) and the other is assigned per flight but does not change the user permission. As from what I understand, the maitre fondu is still a crew member and hence does need specific permission.

Plouc314 (commented):
Overall good, especially the `CalendarModel`.

Few more things:
* I would prefer the package names to be:
  ```
  models
  ├── calendar
  ├── flight
  └── user
  ```
  instead of 
  ```
  dataModels
  ├── calendarModels
  ├── flightModels
  └── userModels
  ```
  it is redundant and I like to keep package names short.
* As `timeSlot` and `date` always go together, we could create a `Date` (or with a better name) class to combine the two
* You should format your code with the `ktfmtFormat` Gradle task

Plouc314:
> Do the user have an attribute role which is either crew, pilot or admin ? If yes, what about more specific roles (maitre fondu) ? Should we do 2 types of role ? One per user (crew, pilot or admin) to determine the permission (user information, ...) and the other is assigned per flight but does not change the user permission. As from what I understand, the maitre fondu is still a crew member and hence does need specific permission.

@Prolas  Each user has an attribute `roleTypes` which is a set of the roles they are allowed to do, see `RoleType` for the possible roles.

RAFInery:
75#issuecomment-2025205513

With @Plouc314  we were quite a bit discussing how to model the user types. Maybe we'll change later on but for the moment we've decided the current model for the following reasons:

The 3 main user types that will use the app are: Pilot, Crew, Admin. They will all have particular screens at some point(e.g. a pilot will have a cockpit view, ...)
The **roles** are more related to the function in one flight and can be seen as different roles a certain user can temporarily assume (or not). We could consider changing the role "crew" to something like "ground" and "pilot" to " captain" to prevent confusion

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/FlightType.kt
```diff
@@ -0,0 +1,27 @@
+package ch.epfl.skysync.dataModels.flightModels
+
+data class FlightType(
+    val name: String,
+    val specialRoles: List<Role> = emptyList()
```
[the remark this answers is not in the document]
RAFInery: test answer...

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/Role.kt
```diff
@@ -0,0 +1,20 @@
+package ch.epfl.skysync.dataModels.flightModels
+
+import ch.epfl.skysync.dataModels.flightModels.RoleType
+import ch.epfl.skysync.dataModels.userModels.User
+
+class Role(
```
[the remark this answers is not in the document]
Plouc314: Sample reply

app/src/main/java/ch/epfl/skysync/dataModels/calendarModels/Availability.kt
```diff
@@ -0,0 +1,17 @@
...
+import java.time.LocalDate
+
+/** Represent the availability of a person for some period */
+data class Availability(
+    val id: String = UNSET_ID,
+    val status: AvailabilityStatus,
+    override val timeSlot: TimeSlot,
+    override val date: LocalDate
+): CalendarViewable {
+    fun setStatus(status: AvailabilityStatus): Availability {
```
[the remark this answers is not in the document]
RAFInery: test answer

app/src/main/java/ch/epfl/skysync/dataModels/userModels/User.kt
Plouc314: Right now you just have duplicated code, as they all have a `roleTypes` attribute anyway, as for the restriction on the roles that can be added, you can just override the default method.

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/Flight.kt
```diff
@@ -0,0 +1,15 @@
+package ch.epfl.skysync.dataModels.flightModels
+
+import java.time.LocalDate
+
+interface Flight{
+    val id: Int
```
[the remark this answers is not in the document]
Plouc314: This is a database-related implementation detail, all stages will have the same `id: String` id. Behind the scene we'll use some relations to have everything consistent, but you shouldn't worry about that here. All this to say, the `flightId` is not needed.

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/FlightType.kt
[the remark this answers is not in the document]
Plouc314: What I mean is, in terms of semantics, it doesn't make sense to have a `Role`, to which a user can be assigned, in the data model that defines a flight type. But I see you've changed it, all good.

app/src/main/java/ch/epfl/skysync/dataModels/calendarModels/CalendarModel.kt
RAFInery: très bonne commentaire, merci

app/src/main/java/ch/epfl/skysync/dataModels/flightModels/Flight.kt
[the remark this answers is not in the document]
RAFInery: got it, thanks

app/src/main/java/ch/epfl/skysync/models/flight/FlightType.kt
```diff
@@ -0,0 +1,23 @@
+package ch.epfl.skysync.models.flight
+
+/**
+ * represents a particular flight type
+ *
+ * @property name name of the type
+ * @property specialRoles roles in addition to the BASE_ROLES(crew, pilot)
+ */
+data class FlightType(val name: String, val specialRoles: List<RoleType> = emptyList()) {
```
Plouc314, line 9: I forgot to mention it on the first review, but `FlightType` also needs an `id` field

app/src/main/java/ch/epfl/skysync/models/user/Admin.kt
```diff
@@ -0,0 +1,23 @@
...
+import ch.epfl.skysync.models.flight.RoleType
+
+class Admin(
+    override val firstname: String,
+    override val lastname: String,
+    override val id: String,
+    override val availabilities: AvailabilityCalendar,
+    override val assignedFlights: FlightGroupCalendar,
+) : User {
+  private val roleTypes: Set<RoleType> = setOf()
```
Plouc314, line 14: I don't really get why `roleTypes` is not part of the `Flight` interface, but if you make the `roleTypes` arrtibute `private` and don't implement the `addRoleType` method, the database side has no way of populating this attribute. (We can leave it as it is for now and change it later, we'll just have to keep it in mind)

app/src/main/java/ch/epfl/skysync/models/calendar/CalendarModel.kt
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  /**
+   * updates the slot by the given coordinates with the output of produceNewValue
+   *
+   * @param date the date coordinate of the slot
+   * @param timeSlot the timeSlot coordinate of the slot
+   * @param produceNewValue computes a new value as function of the coordinates and the old value
+   */
+  fun setByDate(
```
Plouc314, line 78: Why keep the `setByDate` name if even the documentation state "*updates* the slot ..." ?

Plouc314 (approved):
LGTM, few minor comments left.

PR #77, opened 2024-03-28
FirestoreDatabase.kt with all the prints replaced by logs

### LINKED ISSUE #66: Explore logging
See how we can use a logger instead of basic print statements.

* See how we can log into a file
* See how we can handle logging levels (debug, info, ...)
* See if some library pollutes the logs

Zep3k:
closes #66 

Replaces all the prints form FirestoreDatabase.kt with loggers. The prints who check errors are replaced with Log.e() and the others with Log.d() for debug, all under the tag "Firebase".


Prolas (approved):
LGTM

Issue #72, opened 2024-03-27
Data model: User DB implementation


Plouc314:
Define the database schema of the user model and the table class for it.

PR #79, opened 2024-03-28
User roles

Prolas:
Each user is assigned a set of roles.
These roles can be (crew, pilot, admin, maitre fondu). A short description of each role can be found in the readme.

The permission are set here https://console.firebase.google.com/u/0/project/skysync-a1160/firestore/rules. As the collections are not created yet, everybody has access to everything.

In the user collection, the userid should be retrieved as follows 

```kotlin
userid = FirebaseAuth.getInstance().currentUser?.uid
user.value.uid
```
From what I understood the firestore rules uses this id to manage permissions. We could also create a collection roles which maps the userid to a set of roles.

The rules will be of the form:

```
rules_version = '2';
service cloud.firestore {
  match /databases/{database}/documents {
    //Create permission for the collection availability
   //Allow reading for the admin and user concerned if signed in
   //Allow write only for admin if signed in
    match /availability/{availabilityid}{
    	allow read: if isSignedIn() && (isOwner() ||  hasAnyRole(["admin"]) ;
        allow write: if isSignedIn() && hasAnyRole(["admin"])  ;
    }
  
    function isSignedIn(){
      	return request.auth != null;
    }
    
    function isOwner(id){
        let availability = request.resource.data
    	return request.auth.uid == availability.person_id;
    }
    
    function hasAnyRole(roles){
    	return get(/databases/$(database)/documents/user/$(request.auth.uid)).data.role.hasAny(roles);
    }
  }
}
```

@RAFInery can you check that the description of the roles are correct ?


Plouc314 (approved):
LGTM, we'll see where to store the roles later (in #72)

PR #84, opened 2024-04-07
Firestore emulator

### LINKED ISSUE #83: Firestore emulator
* Setup a [Firestore emulator](https://firebase.google.com/docs/emulator-suite/connect_firestore?_gl=1*1daj92i*_up*MQ..*_ga*MTQwMTc0MTY3Mi4xNzEyNDAwNTk1*_ga_CW55HF8NVT*MTcxMjQwMDU5NC4xLjAuMTcxMjQwMDU5NC4wLjAuMA..#kotlin+ktx)
* Refactore the `Database`, `FirestoreDatabase`, `MockDatabase` architecture: remove the mock, setup the `FirestoreDatabase` to connect to the emulator during the tests
* Update the CI to integrate the emulator

Plouc314:
* Setup Firebase emulators
* Refactore `Database`, `MockDatabase`, `FirestoreDatabase`, to only keep `FirestoreDatabase` as we use the emulator for testing, we didn't opt for mock based testing here because it would be very hard to correctly test the query system of Firestore with mocks.
* Refactore the Availability DB implementation to adapt to the `FirestoreDatabase`/emulators architecture
* Update the Readme to document the Firebase emulators

Close #83


app/src/androidTest/java/ch/epfl/skysync/database/tables/AvailabilityTableUnitTest.kt
RAFInery, line 1: how do we distinguish the folders **/androidTest** and **/test**? I understood them such that all non-UI related tests go into **/test**?

app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
```diff
@@ -2,34 +2,72 @@ package ch.epfl.skysync.database
...
+ */
+class FirestoreDatabase(private val useEmulator: Boolean = false) {
   private val db = Firebase.firestore
   private val TAG = "Firebase"
 
-  override fun <T : Any> add(
+  init {
+    if (useEmulator) {
+      val firestore = Firebase.firestore
+      firestore.useEmulator("10.0.2.2", 5002)
```
RAFInery, line 24: I'm maybe not enough familiar with the emulator but did you check if it also runs fine on the remote github machine? Since it will also be used for automatic testing

firestore.rules
RAFInery, line 1: do the rules apply for both the emulator and the real version?

RAFInery (approved):
Good idea to change to the emulator (especially for testing). The advantage of the mockdatabase wast that it was bit more intuitive (at least for me) but I agree that the maintenance would be hard and the emulator is way more elegant

app/src/androidTest/java/ch/epfl/skysync/database/tables/AvailabilityTableUnitTest.kt
Plouc314: My understanding is that **/androidTest** contains all the tests that require running the android emulator (which is required for using the Firestore SDK from what I tested), but maybe we can also put those test in **/test** even if they require the Android emulator ? @Prolas Do you have an idea ?

app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
Plouc314: Yes I setup the CI to start the emulators (@Prolas correct me if I'm wrong)

firestore.rules
Plouc314: It only applies to the emulator, the production database rules are defined [here](https://console.firebase.google.com/u/1/project/skysync-a1160/firestore/databases/-default-/rules), but we ideally want the two to be in sync

app/src/androidTest/java/ch/epfl/skysync/database/tables/AvailabilityTableUnitTest.kt
RAFInery: makes sense, thanks for clearifying

app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
Prolas: Yes, the emulator is launched before the androitTests are executed

Issue #85, opened 2024-04-07
Data model: Flight DB implementation


Plouc314:
Define the database schemas of the flight-related models and the table classes for them.

PR #86, opened 2024-04-07
User database implementation

### LINKED ISSUE #72: Data model: User DB implementation

Plouc314:
* Create `Table` abstract class, refactore `AvailabilityTable`. Avoid some boilerplate for simple operations.
* Create `UserSchema`, `UserTable` classes
* `assignedFlights` is not handled until #85 is implemented
* `roleTypes` is not handled as it is a private attribute at the moment
* The test setup will be simplified in (#106)
* The `SystemClock.sleep` is not an ideal solution, we can improve on it later

Close #72


app/src/main/java/ch/epfl/skysync/models/calendar/CalendarModel.kt
```diff
@@ -109,4 +110,22 @@ abstract class CalendarModel<T : CalendarViewable> {
   fun getByDate(date: LocalDate, timeSlot: TimeSlot): T? {
     return cells.firstOrNull { it.date == date && it.timeSlot == timeSlot }
   }
+
+  /** Return a set of the cells, sorted by date and time slot */
+  private fun getSortedSetOfCells(): SortedSet<T> {
```
RAFInery, line 115: we might consider at some point to directly implement and maintain the underlying list in `CalendarModel` as a sorted structure

app/src/main/java/ch/epfl/skysync/database/Table.kt
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Delete the table
+   *
+   * This is only used for testing, as such it is only supported if using the emulator.
+   *
+   * @param onError Callback called when an error occurs
+   */
+  open fun deleteTable(onError: (Exception) -> Unit) {
```
RAFInery, line 70: maybe define in a separate subclass for the emulator?

app/src/main/java/ch/epfl/skysync/database/DelayedCallback.kt
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This class allows for scheduling a callback function to be executed after a specified delay
+ * period. It provides functionality to decrement the delay period on each invocation of the `run()`
+ * method until the delay reaches 0, at which point the callback function is executed.
+ *
+ * @property delay The initial delay period before executing the callback, specified in units of
+ *   invocation cycles.
+ * @property callback The callback function to be executed after the delay period.
+ */
+class DelayedCallback(private var delay: Int, private val callback: () -> Unit) {
```
RAFInery, line 15: I didn't really understand the use of DelayedCallback. Why did you choose this approach and generally, how do you choose the appropriate delay?

RAFInery (approved):
Seems fine. DB interactions are nicely encapsulated and appear to cover already a vast amount of queries we'll need

app/src/main/java/ch/epfl/skysync/database/DelayedCallback.kt
Plouc314: At the moment I use the callback approach to asynchronous functions, as from what I looked the "async/await" support of Kotlin is pretty poor, the delayed callback is used when launching multiple asynchronous functions "in parallel", as each function takes a callback, but we want to execute the "final" callback only once after all functions are done. You can find an example of utilization in UserTable.deleteAvailabilities`

RAFInery: got  it. Maybe update the description to something like:

_If n similar asynch. operations are executed (in a random order) but the call function should be called only once by the last of the n operations that finishes _ and change the signature to

`DelayedCallback(private var delay: Int, private val callback: () -> Unit)` => 
`ExecuteCallbackOnceByLastOperation(private var nParallelOperations: Int, private val callback: () -> Unit)`

Plouc314: Renamed it to `ParallelOperationsEndCallback`

PR #96, opened 2024-04-08
Minor fix Readme and add missing dependencies

Prolas:
Fixed minor error in Readme
Enable test coverage for jacoco
Added missing dependencies for mock and intents


Plouc314 (approved):
LGTM

PR #106, opened 2024-04-10
Flight database implementation

### LINKED ISSUE #85: Data model: Flight DB implementation

Plouc314:
* Create schemas/tables for: `Balloon`, `Basket`, `Vehicle`, `FlightType`, `Flight`
* Create relation schema/table `FlightMember` to link users to flights (more details in #108)
* Create `DatabaseSetup` class to reduce tests boilerplate, always test with the same database setup
* Need to find a way to handle `null` value when getting multiple items (that is, what if we only receive 3 out of 7 requested items), do we want to fail the entire operation, succeed and log a warning, succeed and display a warning message on the UI

Close #85


mathieumognol (approved):
I learned a lot about how the database works. I didn't see any noticeable problems.

PR #107, opened 2024-04-10
Rules for user and availabilities

### LINKED ISSUE #98: Define role access in database

Prolas:
Collection User:
Each user can read his own entry (see his personal information).
Admin can read/write.

Collection Availability:
Each user can read/write his own availability.
Admin can can read/write.

As the collections are not set up yet these rules are currently ignored as everybody can access everything.

close #98

### DIFF
--- a/firestore.rules
+++ b/firestore.rules
@@ -2,18 +2,47 @@ rules_version = '2';
 
 service cloud.firestore {
   match /databases/{database}/documents {
+  	function isSignedIn(){
+    	return request.auth != null;
+    }
 
-    // This rule allows anyone with your Firestore database reference to view, edit,
-    // and delete all data in your Firestore database. It is useful for getting
-    // started, but it is configured to expire after 30 days because it
-    // leaves your app open to attackers. At that time, all client
-    // requests to your Firestore database will be denied.
-    //
-    // Make sure to write security rules for your app before that time, or else
-    // all client requests to your Firestore database will be denied until you Update
-    // your rules
-    match /{document=**} {
-      allow read, write: if request.time < timestamp.date(2024, 4, 24);
+    function resourceExists(rsc){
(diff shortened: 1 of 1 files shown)


firestore.rules
```diff
@@ -1,19 +1,54 @@
...
-    // leaves your app open to attackers. At that time, all client
-    // requests to your Firestore database will be denied.
-    //
-    // Make sure to write security rules for your app before that time, or else
-    // all client requests to your Firestore database will be denied until you Update
-    // your rules
-    match /{document=**} {
-      allow read, write: if request.time < timestamp.date(2024, 4, 24);
+    	// Allow read, write if document exists and the user is owner or admin
+    	allow read, write: if isSignedIn() && resourceExists(resource) &&
```
mathieumognol, line 50: What's the difference with lines 35 & 37?

firestore.rules
```diff
@@ -1,19 +1,54 @@
...
+    function isAdmin(){
+    	return isOneOfRoles(["admin"])
+    }
+
+    match /user/{userId}{
+      // Allow read if user is owner
+    	allow read: if isSignedIn() && resourceExists(resource) && request.auth.uid == userId;
+      // Allow write if user is admin
+      allow write: if isSignedIn() && resourceExists(resource) &&
+      											(request.auth.uid == userId || isAdmin());
```
mathieumognol, line 38: Why put both "(request.auth.uid == userId || isAdmin())" and not just "isAdmin()"?

mathieumognol (commented):
Overall great rules to protect the database and give the good rights

firestore.rules
Prolas: This is indeed an error. I've also updated the rule so that the administrator has read access.

mathieumognol (approved):

PR #112, opened 2024-04-10
created a user view model

### LINKED ISSUE #94: create first ViewModel
create a ViewModel that holds the states for flights and availabilities

- decide on implementation
- verify how the availabilities and flights calendar as well as the home screen interact through the view model -> define access methods

RAFInery:
closes #94

- after comparing different strategies:  for interactions with the personal calendar (availabilities and assigned flights) there will be a straight forward VM that simply contains the user object

- since the DB is not yet connected, a random user is created on init of the VM

- for other processes of the application new VM will have to be added

- little bugfix: setting the value(s) of a non-existing calendar entry will simply add new entry with the given value(s) i.e. a calendar does no longer have to init the entries with default values


app/src/main/java/ch/epfl/skysync/viewmodel/UserViewModel.kt
```diff
@@ -0,0 +1,57 @@
...
+   * (hardcoded) user.
+   */
+  private fun initUser(firebaseUser: FirebaseUser?) {
+    val foundUser: User? = null // TODO: fetch user from db
+    user = mutableStateOf(foundUser ?: getHardcodeUser())
+  }
+}
+
+/** Returns a hardcoded user till synched with DB */
+fun getHardcodeUser(): User {
```
hacknstack, line 52: Small detail but I don't see the point of doing this method when you could just initiate it at line 47 ( since you commented in 41-44 that you currently just hard-code an user

hacknstack (approved):
LGTM, tests covers the uses of viewmodels quite well and all the methods usefull for the class are implemented, to merge for the advancement of others dependants branches

app/src/main/java/ch/epfl/skysync/viewmodel/UserViewModel.kt
RAFInery: it is just a temporary function that simulates the behaviour of the future function that will return an actual user from the DB. Like this it can quickly be switched out once ready

PR #113, opened 2024-04-10
integrate viewModel into CalendarUi

### LINKED ISSUE #88: Connect Calendar and ModelView
Make sure that the Calendar and the ModelView that are coded independently work together

RAFInery:
closes #88 

- (user) viewModel is separately initialised in the navigator for the personal availabilities calendar and the personal flight calendar. To simplify the architecture the viewmodels were not implemented as shared VM. In case this should cause performance issues  (e.g. due to too big backstack loads) it can still be considered switching

- organised the calendar navigation in a sub-navigation graph

- no specialised accessors for the`UserViewModel` were added => the calendars are read/write to through the existing accessor methods of the calendars

- the two personal calendar screens were integrated

- the program not yet being connected to some persistent DB (firebase or local DB), the changes are only stored temporarily


app/src/main/java/ch/epfl/skysync/navigation/MainGraph.kt
```diff
@@ -17,9 +17,10 @@ fun MainGraph(
     user: FirebaseUser?
 ) {
   NavHost(
-      navController = navHostController,
-      startDestination = if (user == null) Route.LOGIN else Route.MAIN) {
-        homeGraph(navHostController)
+      navController = navHostController, startDestination = Route.MAIN
+      // startDestination = if (user == null) Route.LOGIN else Route.MAIN
```
Zep3k, line 21: Shouldn't you remove the startDestination of line 20 and uncomment line 21

Zep3k (approved):
It's nice that he improved the code of the UI screens

app/src/main/java/ch/epfl/skysync/navigation/MainGraph.kt
RAFInery: the Login still has some problems. It will be switched back once it gets fixed.

PR #114, opened 2024-04-10
Flight calendar

### LINKED ISSUE #99: UI: Flight calendar screen
Create and test the screen of the calendar of the flights

Zep3k:
Screenshot_20240411-103328_SkySync

Close #99


app/src/main/java/ch/epfl/skysync/screens/FlightCalendar.kt
```diff
@@ -0,0 +1,232 @@
...
+ *   within the app.
+ */
+@Composable
+fun ShowFlightCalendar(navController: NavHostController, viewModel: UserViewModel) {
+  Scaffold(bottomBar = { BottomBar(navController) }) { padding ->
+    Box() { Calendar(navController, viewModel, padding) }
+  }
+}
+
+/*
```
Baribal02, line 100: Shouldn't you remove ShowFlightCalendarPreview() as it is commented ?

Baribal02 (approved):
The AddFlight.kt and AddFlightTest.kt will be replaced by the version in #119. We should make sure during the Route is the same and change it if needed

PR #117, opened 2024-04-11
handle non existing entries in AvailabilityCalendar

### LINKED ISSUE #116: handle empty entries in the AvailabilityCalendar
How to handle non existing entries in AvailabilityCalendar?

- Pre-generating entries with default values?

- Implement a strategy

RAFInery:
closes #116 

the AvailabilityCalendar now implements following strat:

- getAvailabilityStatus() returns the AvailabilityStatus of the found entry else UNDEFINED (also an AvailabilityStatus) if not present (instead of null)
- nextAvailabilityStatus() returns next status (in round-robin fashion). 
    -  If the entry is not yet present it initializes(adds the entry) with AvailabilityStatus = OK. 
    - If the current status == NO, then it returns UNDEFINED and removes the entry
 
the round-robin was updated to:
UNDEFINED -> OK -> MAYBE -> NO -> UNDEFINED

Before the null value in the case of a non-existing entry had to be considered separately which is no longer the case


hacknstack (approved):
Great and more consistent way to handle the empty case. LGTM

PR #119, opened 2024-04-11
Add flight form

Baribal02:
The form to add a new flight. 
It is possible to add or delete some vehicles and crew members in the form.
There is some error messages when we try to create a flight without the mandatory fields.
I am not sure about the integration of the delete button in the form as they might be to close of other buttons...
image


app/src/main/java/ch/epfl/skysync/models/flight/FlightType.kt
```diff
@@ -21,3 +28,10 @@ val BASE_ROLES =
         RoleType.PILOT,
         RoleType.CREW,
     )
+
```
Baribal02, line 31: This would be a possible way to do instead to have all_flights in the companion

app/src/main/java/ch/epfl/skysync/models/flight/Flight.kt
```diff
@@ -12,5 +12,5 @@ interface Flight {
   val date: LocalDate
   val timeSlot: TimeSlot
   val vehicles: List<Vehicle>
-  val id: String
+  val id: String?
```
saraanej, line 15: Why can the id of a flight be null ? Don't forget to ensure that appropriate null checks are added for code use and tests as well

app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
```diff
@@ -4,17 +4,54 @@ import androidx.navigation.NavGraphBuilder
...
+import ch.epfl.skysync.screens.AddFlightScreen
 import ch.epfl.skysync.screens.CalendarScreen
 import ch.epfl.skysync.screens.ChatScreen
 import ch.epfl.skysync.screens.FlightScreen
 import ch.epfl.skysync.screens.HomeScreen
+import java.time.LocalDate
 
 /** Graph of the main screens of the app */
 fun NavGraphBuilder.homeGraph(navController: NavHostController) {
+  val listFlights =
```
saraanej, line 20: You should add a comment precising that this is for preview purposes only so that you can get back to it easily when integrating the backend :)

app/src/main/java/ch/epfl/skysync/navigation/MainGraph.kt
```diff
@@ -18,7 +18,7 @@ fun MainGraph(
 ) {
   NavHost(
       navController = navHostController,
-      startDestination = if (user == null) Route.LOGIN else Route.MAIN) {
+      startDestination = /*if (user == null) Route.LOGIN else */ Route.MAIN) {
```
saraanej, line 21: Is the null check for the user performed elsewhere ?

saraanej (commented):
Good code, good job ! A little remark would be to add comments to make code reviews easier for the teammates. Nevertheless, it was quite clear this time :)

app/src/main/java/ch/epfl/skysync/models/flight/Flight.kt
Baribal02: This id is the id that we will get from firebase and with my knowledge about it. I don't think we can get it before writing it in the database. But as we do not have implemented the model view for this part yet so I am not sure about it. It was also easier for me as I did not have to generate random ids for testing it.

app/src/main/java/ch/epfl/skysync/navigation/MainGraph.kt
Baribal02: No, it was just to bypass when I tested it manually and I forgot to put it back

mathieumognol (approved):
Everything seems well implemented for now, while waiting for the viewModel to be available

Issue #159, opened 2024-04-14
Create success/warning/error snackbar


Plouc314:
Create a Snackbar component that could be used from (ideally) anywhere in the application to display, for example, error message when something goes wrong with the database, or when a flight is successfully saved.

Issue #150, opened 2024-04-12
UI: Detail flight information screen


Plouc314:
* Display the details of a flight
* With a `modify` button
* With a `confirm` button

Issue #133, opened 2024-04-12
DB update method


Plouc314:
* Create an `update` method in `FirestoreDatabase`, it updates an existing item with new values
* Create an `update` method in `Table` and children

PR #160, opened 2024-04-14
Viewmodel & Calendar rework

### LINKED ISSUE #129: Connect database to view model
connect the following objects to firebase:

* Get personal flights
* Get (personal) availabilities
* Get all flights (admin) 
* Add availabilities

Plouc314:
Refactored the Calendar screen, reworked the `UserViewModel -> CalendarViewModel`, connected it to the database

* `CalendarUiState.flightGroupCalendar` is not populated yet, we first need to populate the `User.assignedFlights` attribute
* Availabilities are not yet updated (just add/delete), dependent on #133
* Display error messages is not implemented (see issue #159)
* Need UI for the `Save` button of the `AvailabilityCalendar`
* `FlightCalendar.onFlightClick` is not functional, dependent on #150

Close #129


Zep3k (changes requested):
Except for the "collectAsStateWithLifecycle()" the code looks goods. It is well documented and well structured.  Both of the calendar have good modularity and it's nice that they have a similar Ui to the old FlightCalendar.

app/src/main/java/ch/epfl/skysync/screens/Calendar.kt
```diff
@@ -3,55 +3,46 @@ package ch.epfl.skysync.screens
...
 
 @Composable
 fun CalendarScreen(
     navController: NavHostController,
     calendarType: String,
-    viewModel: UserViewModel
+    viewModel: CalendarViewModel
 ) {
   Scaffold(modifier = Modifier.fillMaxSize(), bottomBar = { BottomBar(navController) }) { padding ->
+    val uiState by viewModel.uiState.collectAsStateWithLifecycle()
```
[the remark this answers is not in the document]
Plouc314: I've added dependencies, so running Gradle Sync should get rid of the error

app/src/main/java/ch/epfl/skysync/models/calendar/CalendarModel.kt
```diff
@@ -43,7 +44,7 @@ abstract class CalendarModel<T : CalendarViewable> {
    * @param timeSlot the timeSlot coordinate of the slot
    * @param produceNewValue computes a new value as function of the coordinates and the old value
    */
-  protected fun setByDate(
```
RAFInery, line 46: why did you change the access writes?

RAFInery (approved):
nice work. I think the new interaction architecture between the VM and the UI is cleaner and separates the two a bit better to reduce dependencies. Let's make sure everybody will be up-to-date

app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
```diff
@@ -0,0 +1,139 @@
...
+          factory =
+              object : ViewModelProvider.Factory {
+                override fun <T : ViewModel> create(modelClass: Class<T>): T {
+                  return CalendarViewModel(firebaseUser, userTable, availabilityTable) as T
+                }
+              })
+    }
+  }
+
+  private val _user: MutableStateFlow<User?> = MutableStateFlow(null)
```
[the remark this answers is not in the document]
Plouc314: I used the same convention as from the example [here](https://github.com/android/architecture-samples/blob/main/app/src/main/java/com/example/android/architecture/blueprints/todoapp/taskdetail/TaskDetailViewModel.kt). But I agree we can remove the `_`.

app/src/main/java/ch/epfl/skysync/models/calendar/CalendarModel.kt
Plouc314: I think I needed to use one of the methods at the some point and I didn't see why it wasn't public, I would need check exactly where.

app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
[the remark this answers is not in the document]
RAFInery: I believe it is more in the case you have a public and private variable with the same name in which the private one gets prefixed with _

app/src/main/java/ch/epfl/skysync/models/calendar/CalendarModel.kt
RAFInery: The inheriting classes have their own setters and getters for maintaining the underlying list. So, no need to access the methods from the abstract class directly. I think protected is fine

Plouc314: I've looked and I don't use it in actually, so I'll revert it back

app/src/main/java/ch/epfl/skysync/models/calendar/AvailabilityCalendar.kt
```diff
@@ -44,4 +52,45 @@ class AvailabilityCalendar : CalendarModel<Availability>() {
...
+  ): List<Pair<CalendarDifferenceType, Availability>> {
+    val differences = mutableListOf<Pair<CalendarDifferenceType, Availability>>()
+    for (otherAvailability in other.cells) {
+      val date = otherAvailability.date
+      val timeSlot = otherAvailability.timeSlot
+      val thisAvailability = getByDate(date, timeSlot)
+      if (otherAvailability == thisAvailability) {
+        continue
+      }
+      if (thisAvailability == null || thisAvailability.status == AvailabilityStatus.UNDEFINED) {
```
RAFInery, line 74: UNDEFINED is a psydo status for the availability i.e. it might be returned by a function to the UI components however it is never saved as a state inside the calendar

app/src/main/java/ch/epfl/skysync/models/calendar/AvailabilityCalendar.kt
```diff
@@ -3,8 +3,16 @@ package ch.epfl.skysync.models.calendar
 import ch.epfl.skysync.models.UNSET_ID
 import java.time.LocalDate
 
+/** Represents the type of a cell difference between two calendar */
+enum class CalendarDifferenceType {
```
RAFInery, line 7: You've implemented a type of logging protocol that keeps track of the local changes before persisting it to the DB. However, I ask myself if it would not be less complicated to just store a list of availabilities and updating the lists by simple remplacement of the whole list...

For a list of availabilities = [1,2,3,4]
If we update two elements of the list to [1', 2',3,4] with the current strategy, after assessing what values have changed, we do two requests to update 1 and 2 individually 

But we might as well overwrite the old list with the new list which would no longer need for every entry in the list to have a proper ID and remove the assessment phase.

In order to limit the size of the list we could have some partitioning of the list by range

app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
```diff
@@ -0,0 +1,143 @@
...
+                user?.assignedFlights ?: FlightGroupCalendar(),
+                loadingCounter > 0)
+          }
+          .stateIn(
+              scope = viewModelScope,
+              started = WhileUiSubscribed,
+              initialValue = CalendarUiState(isLoading = true))
+
+  init {
+    uid = firebaseUser.uid
```
RAFInery, line 79: uid could be passed directly as argument since firebaseUser is only used as a container for the uid?

app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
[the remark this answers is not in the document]
Plouc314: In this case I think I just created a new file because it was quite different, otherwise you would see the change if I'm not mistaken

app/src/main/java/ch/epfl/skysync/models/calendar/AvailabilityCalendar.kt
Plouc314: I never make such assumptions unless explicitly stated, but we can state it as such.

app/src/main/java/ch/epfl/skysync/models/calendar/AvailabilityCalendar.kt
Plouc314: This could be a tempting solution, except that in addition to the scalability problem, it makes querying availability more difficult (think of the admin use case), even more so if we start partitioning the list.
So I think the current solution is simpler and more versatile (also think that it'll only be a few dozens requests at most).
I already looked if there was a way to batch the requests but didn't find anything, I'll look again when I have the time.

app/src/main/java/ch/epfl/skysync/models/calendar/AvailabilityCalendar.kt
```diff
@@ -44,4 +52,45 @@ class AvailabilityCalendar : CalendarModel<Availability>() {
...
+      }
+      if (thisAvailability == null || thisAvailability.status == AvailabilityStatus.UNDEFINED) {
+        differences.add(Pair(CalendarDifferenceType.ADDED, otherAvailability))
+      } else if (otherAvailability.status == AvailabilityStatus.UNDEFINED) {
+        differences.add(Pair(CalendarDifferenceType.DELETED, thisAvailability))
+      } else {
+        differences.add(Pair(CalendarDifferenceType.UPDATED, otherAvailability))
+      }
+    }
+    for (thisAvailability in cells) {
```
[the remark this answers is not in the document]
RAFInery: You're right. But it would be more efficient

app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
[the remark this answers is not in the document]
RAFInery: ok

app/src/main/java/ch/epfl/skysync/models/calendar/AvailabilityCalendar.kt
RAFInery: What kind of scalability problem do you see? Each array could be max. a month of availabilities <= 2*31 entries in the array. 

The availabilities could then be modeled:
month: 10
year: 2023
userid: someID
array = (ok, ok, maybe, ...)

Plouc314: Sure, it is something we can keep in mind, but at the moment, unless we notice that the current setup doesn't work. I advise to stick to what we have and move forward

Zep3k (approved):

PR #162, opened 2024-04-15
DB set method

### LINKED ISSUE #131: DB set method
* Add a `set` method to `FirestoreDatabase` that add an item to the database with a specific ID instead of an auto-generated one
* Add a `set` method to `UserTable` to create users with their uid (coming from Firebase authentication)

Prolas:
Created setItem method in FirestoreDatabase.kt
If an id is given the item is set at this id otherwise a new id is generated (same as add)
Modified UserTableUnitTest to test it

close #131

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
+++ b/app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
@@ -63,6 +63,40 @@ class FirestoreDatabase(private val useEmulator: Boolean = false) {
   }
 
   /**
+   * Set a new item to the database
+   *
+   * Set item at id (if null a new id is generated) and override any previously set id.
+   *
+   * @param path A filesystem-like path that specify the location of the table
+   * @param item The item to set to the database, the types of the attributes have to be Firestore
+   * @param id the id of the item (if null a new id is generated)
+   * @param onCompletion Callback called on completion of the operation
+   * @param onError Callback called when an error occurs
+   */
+  fun <T : Any> setItem(
+      path: String,
+      id: String?,
+      item: T,
+      onCompletion: (id: String) -> Unit,
+      onError: (Exception) -> Unit
+  ) {
+    if (id != null) {
+      db.collection(path)
+          .add(item)
+          .addOnSuccessListener { documentSnapshot ->
+            Log.d(TAG, "Created $path/${documentSnapshot.id}")
+            onCompletion(documentSnapshot.id)
+          }
+          .addOnFailureListener { exception ->
+            Log.e(TAG, "Error creating document: ", exception)
+            onError(exception)
+          }
+    } else {
+      addItem(path, item, onCompletion, onError)
+    }
+  }
+
+  /**
    * Get an item from the database
    *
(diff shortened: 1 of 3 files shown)


app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
```diff
@@ -62,6 +62,40 @@ class FirestoreDatabase(private val useEmulator: Boolean = false) {
...
+   *
+   * @param path A filesystem-like path that specify the location of the table
+   * @param item The item to set to the database, the types of the attributes have to be Firestore
+   * @param id the id of the item (if null a new id is generated)
+   * @param onCompletion Callback called on completion of the operation
+   * @param onError Callback called when an error occurs
+   */
+  fun <T : Any> setItem(
+      path: String,
+      id: String?,
```
Plouc314, line 78: In the method `setItem`, you expect to receive an `id`, so you should mark the `id` argument as non-nullable (that way you don't find yourself calling `additem`)

app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
```diff
@@ -62,6 +62,40 @@ class FirestoreDatabase(private val useEmulator: Boolean = false) {
...
+      item: T,
+      onCompletion: (id: String) -> Unit,
+      onError: (Exception) -> Unit
+  ) {
+    if (id != null) {
+      db.collection(path)
+          .add(item)
```
Plouc314, line 85: Pretty sure it should not be `add` (in your whole method you never use the `id` argument so there definitely something fishy...)

app/src/main/java/ch/epfl/skysync/database/tables/UserTable.kt
```diff
@@ -97,6 +97,25 @@ class UserTable(db: FirestoreDatabase) : Table<User, UserSchema>(db, UserSchema:
...
+   *
+   * Set item at id (if null a new id is generated) and override any previously set id.
+   *
+   * @param item The user to add to the database
+   * @param id the id of the item (if null a new id is generated)
+   * @param onCompletion Callback called on completion of the operation
+   * @param onError Callback called when an error occurs
+   */
+  fun set(
+      id: String?,
```
Plouc314, line 111: Same comment as above, `id` should be non-nullable

Plouc314 (changes requested):
Small oversight mistakes but you got the main idea of the database

app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
Prolas: Indeed this should be .document(id).set(item)

app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
Prolas: OK, thanks. I was wondering if the id could be null, and if so, I didn't see the difference with add.

Plouc314 (approved):
LGTM

PR #163, opened 2024-04-16
ConfirmedFlight model

### LINKED ISSUE #143: Create ConfirmedFlight model
* Create model
* Create schema [postponed]
* Integrate in database [postponed]

RAFInery:
closes #143 

## added model

- added `ConfirmedFlight` model and `FlightColor`

- added a method to create a `ConfirmedFlight` from a `PlannedFlight`. The `confirmFlight` method assumes that the flight is ready to be confirmed. Else, throws an exception. For the UI use the function `readyToBeConfirmed` to check if ready before initiating the flight confirmation.

## added functionality for PlannedFlights
- add a roles by a RoleType with `addRoles()`

##

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/models/flight/PlannedFlight.kt
+++ b/app/src/main/java/ch/epfl/skysync/models/flight/PlannedFlight.kt
@@ -7,8 +7,10 @@ import java.time.LocalTime
 data class PlannedFlight(
     override val id: String,
     override val nPassengers: Int,
-    override val team: Team = Team(Role.initRoles(BASE_ROLES)),
     override val flightType: FlightType,
+    override val team: Team = Team(listOf())
+        .addRolesFromRoleType(BASE_ROLES)
+        .addRolesFromRoleType(flightType.specialRoles),
     override val balloon: Balloon?,
     override val basket: Basket?,
     override val date: LocalDate,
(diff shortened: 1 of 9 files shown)


app/src/main/java/ch/epfl/skysync/models/flight/PlannedFlight.kt
```diff
@@ -15,6 +16,44 @@ data class PlannedFlight(
...
+        departureTimeTeam: LocalTime,
+        meetupTimePassenger: LocalTime,
+        meetupLocationPassenger: String,
+        remarks: List<String>,
+        color: FlightColor
+
+
+    ): ConfirmedFlight {
+        if (!readyToBeConfirmed()) {
+            notReadyToConfirm()
```
Plouc314, line 38: I'm not convinced by the `notReadyToConfirm` callback, instead just call `readyToBeConfirmed` before:
```kotlin
if (flight.readyToBeConfirmed()) {
    val confirmedFlight = flight.confirmFlight()
}
```
And throw an exception if `confirmFlight` is called on an not ready to be confirmed flight.
(Also, at the moment you don't return after executing the callback, so you still return a confirmed flight even if it is not ready to)

RAFInery: yep, thanks, you were too quick:-) I've replaced it with an exception

Plouc314 (approved):
LGTM

PR #165, opened 2024-04-16
add flight status to flights

### LINKED ISSUE #155: add status to flights
- add necessary model to flights
- add methods to update status

status: 
- planned (if flight was created but not yet all details have been filled in)
- ready (flight has all details but is not yet confirmed)
- confirmed (has been confirmed)

### LINKED ISSUE #157: UI: add flight status to overview
- think as to where to add the status
- add the status to the UI

RAFInery:
closes #157 
closes #155 

## added FlightStatus
- `FlightStatus` added as enum with a display name and display color
- `Flight` interface extended with `getFlightStatus()` that returns the current status

## Integrated FlightStatus into UI of overview

- the flight status is displayed as a little label and with a dedicated light color

Screenshot 2024-04-16 at 14 30 44


app/src/main/java/ch/epfl/skysync/models/flight/FlightStatus.kt
RAFInery, line 1: @Plouc314 I've added a more granular status for the Flights to be displayed to the user. For now they share the same name with the already existing FlightStatus from the DB but we might change one of the two names to prevent any confusion. The way I understand it they are a bit different:

FlightStatus (DB): helps to determine the table/schema

FlightStatus(in app): helps to display the current state of the flight to the user

As an example: the states **ready** and **planned** tell the admin whether or not a flight is ready to be confirmed however from a DB point of view they're the same...
Any comments?

PR #171, opened 2024-04-18
Viewmodel for flight overview

RAFInery:
## new ViewModel for the Flights overview
- added an viewmodel that handles accesses and modifications of all current flight related UI
- > (home screen, add flight, modify flight, flight details)

- currently synchronisation errors with the DB are not handled however the local state of the VM might still be updated => undefined behaviour


app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
```diff
@@ -0,0 +1,166 @@
...
+
+    flightTable.getAll(
+        { flights -> _currentFlights.value = flights },
+        { exception -> Log.d("FLightrefresh", exception.toString()) })
+  }
+
+  /**
+   * modifies the flight by deleting the old flight and adding a new one in the db and the viewmodel
+   */
+  fun modifyFlight(
```
Prolas, line 117: There is a function update in FlightTable since the merge of #176 which suits well here

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+  @Before
+  fun setUp() {
+    dbSetup.clearDatabase(db)
+    dbSetup.fillDatabase(db)
+    composeTestRule.setContent {
+      viewModel =
+          FlightsViewModel.createViewModel(
+              flightTable, balloonTable, basketTable, flightTypeTable, vehicleTable)
+
+      // val flights = viewModel.currentFlights.collectAsStateWithLifecycle()
```
Prolas, line 48: Leftover comments

Prolas:
Why did you create a second DatabaseSetup ?

RAFInery:
I've created the second DBsetup not to interfere with other tests

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+      // Text(flights.value.getOrNull(0)?.toString()?: "No flight types")
+    }
+  }
+
+  @Test
+  fun fetchesCurrentFlightsIfEmpty() {
```
Prolas, line 54: Nitpick: I prefer fetchCurrentFlightsIfEmpty

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+            timeSlot = TimeSlot.AM,
+            vehicles = listOf(dbSetup.vehicle1),
+            id = UNSET_ID)
+
+    viewModel.addFlight(flight1)
+    SystemClock.sleep(DB_SLEEP_TIME)
+
+    viewModel.refreshCurrentFlights()
+    SystemClock.sleep(DB_SLEEP_TIME)
+    Assert.assertEquals(viewModel.currentFlights.value.size, 1)
```
Prolas, line 130: The DB is note reset before each test. Hence this test fails as there are 3 flights (2 from previous test and the new added)

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+
+    flightTable.add(flight1, { flight1 = flight1.copy(it) }, { error -> println(error) })
+
+    flightTable.add(flight2, { flight2 = flight2.copy(it) }, { error -> println(error) })
+
+    SystemClock.sleep(DB_SLEEP_TIME)
+
+    viewModel.refreshCurrentFlights()
+    SystemClock.sleep(DB_SLEEP_TIME)
+    Assert.assertEquals(viewModel.currentFlights.value.size, 2)
```
Prolas, line 103: Assertions work the other way around. At the left there is the expected value, and on the right the actual value.

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
RAFInery: there is a clearDatabse() in the @Begin of the test function

Prolas:
> I've created the second DBsetup not to interfere with other tests

I understand the idea but you could use the same database and then add your own flights.

Prolas:
For the CI I do not really understand why it fails but in main the function add in userTable is replaced with set. You can merge with main and replace the lines pointed out by the CI  (113 to 116 in DatabaseSetup2) with userTable.set.

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
```diff
@@ -0,0 +1,212 @@
...
+  private val flightTypeTable = FlightTypeTable(db)
+  private val vehicleTable = VehicleTable(db)
+
+  @get:Rule val composeTestRule = createComposeRule()
+  lateinit var viewModel: FlightsViewModel
+
+  @Before
+  fun setUp() {
+    dbSetup.clearDatabase(db)
+    dbSetup.fillDatabase(db)
```
RAFInery, line 42: @Prolas here the DB is set up

RAFInery:
I've combined the setup into one Class and added a new setup function

RAFInery:
I had to merge the incoming changes from main into my branch before pushing

Prolas (approved):
LGTM but in the modifyFlight method it would be better to use the update method from FlightTable

PR #176, opened 2024-04-18
DB update method

### LINKED ISSUE #133: DB update method

Prolas:
close #133 
For now the update method is done and tested for the UserTable, AvailabiltiyTable and FlightTable. 
It sets the updated item at the given id (id does not change on update).
For FlightTable it additionally updates the FlightMember table.


Plouc314 (approved):
I refactored the UserTable tests in #174, so you will have conflicts (you can add more things in DatabaseSetup as needed).
Otherwise good

PR #189, opened 2024-04-19
Snackbar component

### LINKED ISSUE #159: Create success/warning/error snackbar

mathieumognol:
Close #159 
Capture d'écran 2024-04-19 214228


Plouc314 (commented):
Overall looks good, I have questions about practical use:

I'm guessing we use `SnackbarManager.showMessage` to display a snackbar message, can we call this function from anywhere in the code (that is not from a UI component or from another thread) ?

In the preview you pass the `GlobalSnackbarHost` to the `Scaffold`, should we create a single `GlobalSnackbarHost` instance and pass it down to each screen or create one for each screen ?

mathieumognol:
To answer plouc's questions: 

1) Since SnackbarManager is a singleton, you can call SnackbarManager.showMessage(message) from anywhere in the app to enqueue messages to be displayed. This can be done from any UI or non-UI component.

2) In the last pr I created a single instance of GlobalSnackbarHost in the app's root Scaffold. This approach allows us to have a global snackbar manager that can display messages from anywhere in your app without needing to pass it down or recreate it in multiple places.

Plouc314:
@Prolas is there a way to have the previews not count in the line coverage ? Otherwise I advise to juste remove the previews completely

Plouc314 (approved):
LGTM

Prolas:
> @Prolas is there a way to have the previews not count in the line coverage ? Otherwise I advise to juste remove the previews completely

Yes, there is a way by adding a our own tag but as the preview function are not very long and should not affect line coverage too much overall.

PR #192, opened 2024-04-21
Simple Timer

Baribal02:
Add of a timer for the flight time added to the FlightScreen
There is only two actions on the timer start and when it has already started reset
image


app/src/main/java/ch/epfl/skysync/ui/components/Timer.kt
```diff
@@ -0,0 +1,73 @@
...
+  Box(
+      modifier = modifier.testTag("Timer"),
+      contentAlignment = androidx.compose.ui.Alignment.Center) {
+        Column(horizontalAlignment = androidx.compose.ui.Alignment.CenterHorizontally) {
+          Text(
+              modifier = Modifier.testTag("Timer Value"),
+              text = timer.formatTime(),
+              style = MaterialTheme.typography.headlineMedium)
+          if (isRunning) {
+            Button(
```
mathieumognol, line 35: Would've been nice to add color = lightOrange to keep the visuals of the other buttons on the screen

mathieumognol (approved):
Apart from a small UI detail, everything seems good to me, especially implementing the timer as a component rather than coding it directly in Flight screen

app/src/main/java/ch/epfl/skysync/ui/components/Timer.kt
Baribal02: It's done

PR #195, opened 2024-04-22
Chat backend

### LINKED ISSUE #180: Explore Chat backend
* Look how to store message (use Firestore ?, if yes how to structure data)
* Look to create Table/logic

Plouc314:
* Create `Message`,  `MessageGroup` schemas/tables
* Create a `queryListener` method to `FirestoreDatabase`, `Table` that setup a listener that reacts to changes in the results of a particular query ([doc](https://firebase.google.com/docs/firestore/query-data/listen#kotlin+ktx))
* Create a `addGroupListener` to `MessageTable` to setup a listener that reacts to messages on a specific message group

Close #180


Prolas:
I will update the database diagram with the new tables

app/src/androidTest/java/ch/epfl/skysync/database/DatabaseSetup.kt
```diff
@@ -164,6 +179,9 @@ class DatabaseSetup {
             launch { basket2 = basket2.copy(id = basketTable.add(basket2)) },
             launch { vehicle1 = vehicle1.copy(id = vehicleTable.add(vehicle1)) },
             launch { vehicle2 = vehicle2.copy(id = vehicleTable.add(vehicle2)) },
+            launch {
+              messageGroup1 = messageGroup1.copy(id = messageGroupTable.add(messageGroup1))
```
Prolas, line 183: Why do you add the messageGroup1 here and not at the end when the the set of user is added ?

app/src/androidTest/java/ch/epfl/skysync/database/tables/MessageGroupTest.kt
```diff
@@ -0,0 +1,100 @@
...
+    dbs.clearDatabase(db)
+    dbs.fillDatabase(db)
+  }
+
+  @Test
+  fun getTest() = runTest {
+    val messageGroup = messageGroupTable.get(dbs.messageGroup1.id, onError = { assertNull(it) })
+    assertNotNull(messageGroup)
+    assertTrue(
+        listOf(dbs.admin2.id, dbs.pilot1.id, dbs.crew1.id).containsAll(messageGroup!!.userIds))
```
Prolas, line 34: This is true even if messageGroup !!.userIds is empty. It would be better if it were the other way around.

app/src/main/java/ch/epfl/skysync/database/tables/MessageGroupTable.kt
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Update a message group
+   *
+   * This will not update the group messages, only the list of users who are part of it.
+   *
+   * @param item The message group to update
+   * @param id The id of the message group
+   * @param onError Callback called when an error occurs
+   */
+  suspend fun update(id: String, item: MessageGroup, onError: ((Exception) -> Unit)? = null) {
```
Prolas, line 104: Nitpick: what do you think of updateGroupUser ?

app/src/main/java/ch/epfl/skysync/models/message/Message.kt
```diff
@@ -0,0 +1,10 @@
+package ch.epfl.skysync.models.message
+
+import ch.epfl.skysync.models.UNSET_ID
+import java.util.Date
+
+data class Message(
```
Prolas, line 6: Why does a message not have an owner ?

Prolas (commented):
Good overall, but a few errors in the tests

app/src/androidTest/java/ch/epfl/skysync/database/DatabaseSetup.kt
Plouc314: Because I need the message group id to add the messages, but good catch I need to set the userIds before adding the message group, will fix

app/src/androidTest/java/ch/epfl/skysync/database/tables/MessageGroupTest.kt
Plouc314: Good catch, will fix

app/src/main/java/ch/epfl/skysync/database/tables/MessageGroupTable.kt
Plouc314: We can go for `updateGroupUsers` as long as it's the only attribute that gets updated

app/src/main/java/ch/epfl/skysync/models/message/Message.kt
Plouc314: Very good question, I have no idea. Will fix

Prolas (approved):
LGTM

PR #200, opened 2024-04-23
Search bar

Baribal02:
Add of a search Bar to be use in forms (for example for users in the flight form)

image


app/src/main/java/ch/epfl/skysync/components/forms/FlightForm.kt
```diff
@@ -592,6 +576,32 @@ fun RoleField(
       }
 }
 
+@Composable
+fun <T> TitledInputTextField(
```
Plouc314, line 580: This could be extracted as a separated component, for ex: `components/TitledInputTextField.kt`

Baribal02: I have done it for this and also for CustomDropDownMenu

app/src/main/java/ch/epfl/skysync/components/forms/SearchBarCustom.kt
```diff
@@ -0,0 +1,117 @@
...
+    onElementClick: (T) -> Unit,
+    propositions: List<T>,
+    showProposition: (T) -> String,
+    placeholder: String = "Search"
+) {
+  val focusRequester = remember { FocusRequester() }
+  var height by remember { mutableStateOf(90.dp) }
+  val animatedHeight by animateDpAsState(targetValue = height, label = "")
+  val keyboardController = LocalSoftwareKeyboardController.current
+  LaunchedEffect(active) { height = if (active) 300.dp else 90.dp }
```
hacknstack, line 50: You could do like the dropdownmenu and make the height modular

hacknstack (approved):
LGTM, I really like how modular the composable are.

app/src/main/java/ch/epfl/skysync/components/forms/SearchBarCustom.kt
Baribal02: The height change is necessary to make the difference between the active and inactive states. Also more than 300.dp when it is opened it not really usefull as there is the keyboard that will mask it and we won't be able to click on it.

PR #203, opened 2024-04-23
Fix User roleTypes

### LINKED ISSUE #164: User fix roleTypes
* Set the `roleTypes` attribute to public
* Take it into account in the db schema

Plouc314:
Set the `roleTypes` attribute in `User` (thus public) instead of previously private attribute defined in each Pilot, Crew, Admin

Close #164


mathieumognol (approved):
LGTM

PR #205, opened 2024-04-23
Add User Screen

Baribal02:
Form to add a user

Fields : 
- User's Role
- First Name
- Last Name
- E-mail
- Balloon qualification if needed

For now it does not add anything as it is not connected with a viewModel to the database

image


Prolas (approved):
LGTM

PR #209, opened 2024-04-24
End to end test

### LINKED ISSUE #184: End-to-end test 2
As an admin, I want to add flights so that crew/pilots can be assigned

hacknstack:
Closes #184 ,adds an end-to-end test for add flight

### DIFF
--- a/app/src/androidTest/java/ch/epfl/skysync/end_to_end/E2EAddFlights.kt
+++ b/app/src/androidTest/java/ch/epfl/skysync/end_to_end/E2EAddFlights.kt
@@ -1,6 +1,5 @@
 package ch.epfl.skysync.end_to_end
 
-import androidx.compose.runtime.mutableStateOf
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.test.hasTestTag
 import androidx.compose.ui.test.junit4.createComposeRule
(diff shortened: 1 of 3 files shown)


Zep3k (changes requested):
Shouldn't you confirm and check if the flight is created

app/src/androidTest/java/ch/epfl/skysync/end_to_end/E2EAddFlights.kt
```diff
@@ -0,0 +1,103 @@
...
+    composeTestRule.onNodeWithTag("Basket Menu").performClick()
+    composeTestRule.onNodeWithTag("Basket 1").performClick()
+    val title1 = "Add Flight"
+    composeTestRule.onNodeWithTag("$title1 Button").performClick()
+    route = navController.currentBackStackEntry?.destination?.route
+    Assert.assertEquals(Route.HOME, route)
+    var flightIsCreated = false
+    runTest {
+      val flights = repository.flightTable.getAll(onError = { assertNull(it) })
+      flightIsCreated = flights.any { it.nPassengers == 13 }
```
Zep3k, line 99: this will cause the test to always pass if there's actually a flight with 13 passengers

Zep3k (approved):

app/src/androidTest/java/ch/epfl/skysync/end_to_end/E2EAddFlights.kt
hacknstack: iirc it is initialised at test run so the db is empty.

PR #232, opened 2024-04-25
Make the calendar great again

### LINKED ISSUE #204: Rework the calendar and its tests

Prolas:
close #204 
Reworked the UI of the Calendar (Availability Calendar and Flight Calendar)
The navigation between the 2 calendars is now on top of the screen
 Replaced navigation buttons between weeks with swipe.

Availability Calendar:
* Addition of cancel and save buttons

The connection with the View Model is still missing. For example, availabilies are clickable but not stored in the database.

image
image


app/src/main/java/ch/epfl/skysync/components/FlightCalendar.kt
```diff
@@ -32,15 +37,13 @@ import java.time.LocalDate
...
 @Composable
 fun FlightTile(date: LocalDate, time: TimeSlot, flight: Flight?, onClick: () -> Unit) {
-  var weight = 0.5f
-  if (time == TimeSlot.PM) {
-    weight = 1f
-  }
+  val testTag = date.toString() + time.toString()
+  val weight = if (time == TimeSlot.AM) 0.5f else 1f
   if (flight != null) {
     Button(
```
mathieumognol, line 43: I wonder if using "Card" rather than "Button" would be more appropriate but anyways it works here

app/src/main/java/ch/epfl/skysync/components/AvailabilityCalendar.kt
```diff
@@ -4,32 +4,36 @@ import androidx.compose.foundation.background
...
 import ch.epfl.skysync.models.calendar.TimeSlot
+import ch.epfl.skysync.ui.theme.*
 import java.time.LocalDate
 
 // Define custom colors to represent availability status
 // These colors are used to indicate availability status in the UI
 
 // Custom color for indicating availability status as "OK" (e.g., available)
-val customGreen = Color(android.graphics.Color.parseColor("#aaee7b"))
+val customGreen = Color(0xffaaee7b)
```
mathieumognol, line 27: What is the difference in usage between this one and the one added in Color?

mathieumognol (approved):
LGTM

app/src/main/java/ch/epfl/skysync/components/AvailabilityCalendar.kt
Prolas: It the the same color but the second solution is shorter

app/src/main/java/ch/epfl/skysync/components/FlightCalendar.kt
Prolas: Yes, it's probably better, but it's from the previous version of the calendar and as it seemed to work, I didn't want to change it.

PR #244, opened 2024-04-26
fix of the drop down bug

Baribal02:
image

The drop down show itself at the right place


Prolas (approved):
LGTM. Beautiful column

PR #253, opened 2024-04-29
Sharing Location between Users

### LINKED ISSUE #215: share location between users
use firestore listener

mathieumognol:
Close #215


app/src/main/java/ch/epfl/skysync/models/location/Location.kt
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Represents a Location
+ *
+ * @param user The user of whom we share the location
+ * @param value His actual location
+ */
+data class Location(
+    val id: String = UNSET_ID,
+    val user: User,
```
Plouc314, line 15: I don't think you need to have a complete user here, `userId` is enough (It is defined that way in `Message` so that we can display the sender name)

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
```diff
@@ -0,0 +1,55 @@
...
+    db: FirestoreDatabase,
+) : Table<Location, LocationSchema>(db, LocationSchema::class, PATH) {
+
+    /**
+     * Update the location of a user in the database.
+     *
+     * @param location The location to update.
+     * @param onError Optional error handler for catching exceptions.
+     */
+    suspend fun updateLocation(location: Location, onError: ((Exception) -> Unit)? = null) {
```
Plouc314, line 23: It's a good idea to always update the location instead of creating a new one each time, however we need to choose what ID to set for the location, a very straightforward solution would be to directly use the user ID (thus removing the `userId` field), as we expect to store only one location for a user at any given time.

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
```diff
@@ -0,0 +1,55 @@
...
+     * @param onChange Callback called each time there is an update.
+     * @param coroutineScope CoroutineScope for launching asynchronous tasks.
+     */
+    fun listenForLocationUpdates(
+        onChange: suspend (List<Location>) -> Unit,
+        coroutineScope: CoroutineScope,
+        onError: ((Exception) -> Unit)? = null
+    ): ListenerRegistration {
+        return db.queryListener(
+            path, Filter.equalTo("active", true), LocationSchema::class, { update ->
```
Plouc314, line 42: Here what is the`"active"` field ? `LocationSchema` doesn't not have such a field, here I would take an `userIds` argument so that the query select which users location to listen to

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param onChange Callback called each time there is an update.
+     * @param coroutineScope CoroutineScope for launching asynchronous tasks.
+     */
+    fun listenForLocationUpdates(
+        onChange: suspend (List<Location>) -> Unit,
+        coroutineScope: CoroutineScope,
+        onError: ((Exception) -> Unit)? = null
+    ): ListenerRegistration {
+        return db.queryListener(
```
Plouc314, line 41: Also you could use the `Table.queryListener` method (which `LocationTable` inherits) instead of directly calling the `FirestoreDatabase.queryListener`

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
```diff
@@ -0,0 +1,55 @@
...
+                    withErrorCallback(onError) {
+                        val locations = update.updates.map { it.toModel() }
+                        onChange(locations)
+                    }
+                }
+            }, coroutineScope)
+    }
+
+    companion object {
+        const val PATH = "locations"
```
Plouc314, line 53: Nitpick: with the current convention, we write collection names on singular, here it would be: `"location"`

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
```diff
@@ -0,0 +1,55 @@
...
+     */
+    fun listenForLocationUpdates(
+        onChange: suspend (List<Location>) -> Unit,
+        coroutineScope: CoroutineScope,
+        onError: ((Exception) -> Unit)? = null
+    ): ListenerRegistration {
+        return db.queryListener(
+            path, Filter.equalTo("active", true), LocationSchema::class, { update ->
+                coroutineScope.launch {
+                    withErrorCallback(onError) {
```
Plouc314, line 44: There is no need for an `withErrorCallback` wrapper here (I put one in `MessageGroupTable.addGroupListener` because I need to fetch information from the database in the callback, which you don't need to do)

app/src/main/java/ch/epfl/skysync/database/schemas/LocationSchema.kt
```diff
@@ -0,0 +1,44 @@
...
+    )
+
+data class LocationSchema(
+    val id: String? = null,
+    val userId: String? = null,
+    val latitude: Double? = null,
+    val longitude: Double? = null
+) : Schema<Location> {
+    override fun toModel(): Location {
+        if (userId == null || latitude == null || longitude == null || id == null) {
```
Plouc314, line 27: This is not needed, the attributes will always be not null except the corresponding attribute in the model is nullable (see `FlightSchema` for an example)

app/src/main/java/ch/epfl/skysync/models/location/Location.kt
mathieumognol: Makes it way simpler indeed

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
mathieumognol: yep

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
mathieumognol: Used the `Table.queryListener` method at the end

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
mathieumognol: Indeed

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
mathieumognol: Good to know

app/src/main/java/ch/epfl/skysync/database/tables/LocationTable.kt
mathieumognol: Noted

app/src/main/java/ch/epfl/skysync/database/schemas/LocationSchema.kt
mathieumognol: Okay

mathieumognol:
It's my first time doing test for ViewModel and database code, I have a problem with the LocationTableTest running indefinetly on my machine and in the LocationViewModelTest, I get error: "After waiting for 10s, the test coroutine is not completing". If anybody has a tip to solve these issues

mathieumognol:
> It's my first time doing test for ViewModel and database code, I have a problem with the LocationTableTest and the LocationViewModelTest, I get error: "After waiting for 10s, the test coroutine is not completing". If anybody has a tip to solve these issues

So the problem doesn't come from my code because I get the same error when running test from other ViewModel/Table

mathieumognol:
Finally I think the test are correct. There must be an error in  'updateLocation' or in 'listenForLocationUpdates' but I can't figure it out.

Plouc314:
> Finally I think the test are correct. There must be an error in `updateLocation` or in `listenForLocationUpdates` but I can't figure it out.

I corrected some issues you can have a look, it should work now

Plouc314 (approved):
LGTM, the `Flight` screen is a bit messy, we might want to refactore at some point

Issue #220, opened 2024-04-25
User creation: find a way to link uid to user


Prolas:
The id of the user in the database is the same as the uid of a firebase auth user, we need to find a way that when an admin create a new user, it is linked with this uid.

One idea could be that when created by an admin, the user is first put in a _partially created_ user table, without its uid (as it is unknown at this point), then when the new user log in the first time, if no user is find for his uid (which will be the case), look into the _partially created_ user table to see if an email address match, then create a new user in the "real" user table.
This has the drawback that any user need the write access for the user table.

PR #254, opened 2024-04-29
bugfix: calendar is not synched for all users

### LINKED ISSUE #245: fix personal calendar DB synchronization
with multiple reworks of the calendar screens the DB sync. was broken along the way

- find the issue
- fix it 
=> on click of the "save" button the changes in the availability calendar should get persisted to the DB

RAFInery:
closes #245

# problem
the changes in the calendar of some users are synchronized with the DB whereas for other users they're not properly synchronized.

# reason
The way we've implemented a user so far every new user must (1) be able to connect to the firebase DB and (2) have an entry in the UserTable. (2) being not yet automatically handled, the calendar synchronized only properly in the case the user has manually added their user in the UserTable for some reason.

# fix
in `userRefresh()` if `userTable.get() `returns null a new Admin user is automatically added to the userTable 

# comments
User that will be added through the app (as it will be implemented in #220) will be properly initialized with (1) and (2)
In any other case, an admin entry in the userTable will be created to simplify testing and usage for the people grading the milestones .


Plouc314 (approved):
Very hacky but it'll do the job

PR #258, opened 2024-04-30
Show user specific flights

### LINKED ISSUE #226: VM: show Crew/pliot specific flights in overview
- add a function `getFlightsFor(u: User)` to the FlightsViewModel that returns the flights for a particular user
- since it depends on #219, do a mock version first and connect to the real DB in a different issue( issue here)

RAFInery:
closes #226

# show flights as a function of User on the overview screen

- Admin sees all flights
- Pilot/Crew sees only the flights his assigned to

# comments
the `User` (along with its UserRole) is loaded in the `FlightsViewModel`. Later, we might choose to do this outside of the ViewModel and directly pass the loaded User to the ViewModels as an argument

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
+++ b/app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
@@ -1,5 +1,6 @@
 package ch.epfl.skysync.viewmodel
 
+import android.util.Log
 import androidx.compose.runtime.Composable
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.ViewModelProvider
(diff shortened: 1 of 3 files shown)


Zep3k (approved):
LGTM

app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
```diff
@@ -45,21 +53,37 @@ class FlightsViewModel(
...
+    refreshUserAndFlights()
     refreshCurrentBalloons()
     refreshCurrentBaskets()
     refreshCurrentFlightTypes()
     refreshCurrentVehicles()
   }
 
+  fun refreshUserAndFlights() =
+      viewModelScope.launch {
+        _currentUser.value = repository.userTable.get(userId ?: UNSET_ID, onError = { onError(it) })
```
Plouc314, line 75: You should not try to fetch a user with an unset id, it will throw an error, instead you should require to have an not null uid in your view model (which will always be the case anyway), so you could have at line 32: ’val uid: String,’

RAFInery: got it, thanks

PR #259, opened 2024-05-01
change timer to continue running on all screens

### LINKED ISSUE #222: Timer provider global
Put a timer provider/shared state in the root to have its state preserved when switching between screens

RAFInery:
closes #222

# started counter continues running when flight screen is left
- added `TimerViewMode` to wrap the timer related information
- the timer viewmodel is initiated on the MainActivity level as a first solution
- > timer must continue running in the back even if we leave the Flight screen. However, the flight screen is the only screen that actually accesses the variables of the `TimerViewModel` hence I decided against a shared viewmodel implementation strategy. 

# increased precision of counter
- on start:  a coroutine updates the counter (miliseconds) with a delay of 100ms
- uses now timestamp based computations instead of simply increasing the timer with delays since the delays are not exact and depend on the scheduler + the current workload
-

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
+++ b/app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
@@ -49,9 +49,7 @@ fun NavGraphBuilder.homeGraph(
           ChatViewModel.createViewModel(uid!!, messageListenerSharedViewModel, repository)
       ChatScreen(navController, chatViewModel)
     }
-    composable(Route.FLIGHT) {
-        FlightScreen(navController, timer!!)
-    }
+    composable(Route.FLIGHT) { FlightScreen(navController, timer!!) }
     composable(Route.HOME) { entry ->
 
       // get the MessageListenerSharedViewModel here so that it gets
--- a/app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
+++ b/app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
@@ -29,12 +29,14 @@ import ch.epfl.skysync.screens.flightDetail.FlightDetailScreen
 import ch.epfl.skysync.viewmodel.ChatViewModel
 import ch.epfl.skysync.viewmodel.FlightsViewModel
 import ch.epfl.skysync.viewmodel.MessageListenerSharedViewModel
+import ch.epfl.skysync.viewmodel.TimerViewModel
 
 /** Graph of the main screens of the app */
 fun NavGraphBuilder.homeGraph(
     repository: Repository,
     navController: NavHostController,
-    uid: String?
+    uid: String?,
+    timer: TimerViewModel
 ) {
   navigation(startDestination = Route.HOME, route = Route.MAIN) {
     personalCalendar(repository, navController, uid)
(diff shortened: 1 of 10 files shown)


app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
```diff
@@ -29,12 +29,14 @@ import ch.epfl.skysync.screens.flightDetail.FlightDetailScreen
...
 import ch.epfl.skysync.viewmodel.MessageListenerSharedViewModel
+import ch.epfl.skysync.viewmodel.TimerViewModel
 
 /** Graph of the main screens of the app */
 fun NavGraphBuilder.homeGraph(
     repository: Repository,
     navController: NavHostController,
-    uid: String?
+    uid: String?,
+    timer: TimerViewModel? = null
```
Plouc314, line 39: Why make timer nullable in `homeGraph` when it is not in `MainGraph` and thus always passed ?

app/src/main/java/ch/epfl/skysync/viewmodel/TimerViewModel.kt
```diff
@@ -0,0 +1,93 @@
...
+  fun stop() {
+    _isRunning.value = false
+    job?.cancel()
+    job = null
+    lastTimestamp = 0L
+  }
+
+  override fun onCleared() {
+    super.onCleared()
+    stop()
```
Plouc314, line 91: It would make more sense to first call `stop` and then call parent `onCleared` to me but perhaps it doesn't make a difference.

Plouc314 (approved):
Looks good, we can go with the defined-at-top-and-then-pass-down approach, but I think we can also use a shared view model and define it only in the screen that actually interest us (i.e. flight) and it will still be kept across all screen of the same navigation graph (see how it is done with `messageListenerSharedViewModel`)

app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
RAFInery: this was just an easy workaround to not change all the tests

app/src/main/java/ch/epfl/skysync/viewmodel/TimerViewModel.kt
RAFInery: I don't think it matters

RAFInery:
The way I understood the sharedViewModel is that we sometimes want to have the same instance of a Viewmodel that will be used throughout multiple navigation destinations/screens. Once none of the screens that share the viewmodel is no longer active the VM is destroyed and rebuild the next time one of those screens comes back online. In our case that would mean that we'd have to init the SharedViewModel for the timer in each navigation destination in order to make sure that the counter is not reset.

app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
Plouc314: Please state it explicitly in a comment then

Plouc314:
> The way I understood the sharedViewModel is that we sometimes want to have the same instance of a Viewmodel that will be used throughout multiple navigation destinations/screens. Once none of the screens that share the viewmodel is no longer active the VM is destroyed and rebuild the next time one of those screens comes back online. In our case that would mean that we'd have to init the SharedViewModel for the timer in each navigation destination in order to make sure that the counter is not reset while we navigate.

I'm afraid that is not correct, please look how it has been done for `MessageListenerSharedViewModel`, you do not have to define it everywhere, one route in the navigation graph is enough

PR #262, opened 2024-05-02
Assign available users and equipment

### LINKED ISSUE #212: VM: assign/choose only available users and equipment
- for a given flight choose only among available users
-> add function to `FlightsViewModel`: `getAvailableUsersFor(date, timeslot): List<User>`
- for a given flight choose only among available equipment:
-> add functions to `FlightsViewModel` for all equipment types e.g. `getAvailableBasketsFor(date, timeslot): List<Basket>`
- inject new functions into `FlightForm`

RAFInery:
closes #212 

# assign users
- users can now get assigned to flights
- the changes are persisted in the DB

# updated ViewModel 
- added availableUsers and available[EquipmentName] to `FlightsViewModel`  as a function of date and time
- if no date and timeslot are chosen the VM returns all users and equipments, else it returns the equipment/users filtered for their respective availability

# changes in the FlightForm UI
- integrated the user assignment into the `FlightForm` UI
- some minor changes in the display
- some modularisation for the Team assignment and extension UI


app/src/main/java/ch/epfl/skysync/models/user/Admin.kt
```diff
@@ -16,4 +16,8 @@ data class Admin(
   override fun addRoleType(roleType: RoleType): Admin {
     return this.copy(roleTypes = roleTypes + roleType)
   }
+
+    override fun toString(): String {
```
Plouc314, line 20: In my opinion overriding the `toString` method to only return the name is misleading, instead you could call it `name`,
also why define it in all class that implements `User` but not in `User` directly ?

app/src/main/java/ch/epfl/skysync/viewmodel/FlightsOverviewViewModel.kt
```diff
@@ -0,0 +1,73 @@
...
+import ch.epfl.skysync.models.user.Pilot
+import ch.epfl.skysync.models.user.User
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.launch
+
+/** ViewModel for the user */
+class FlightsOverviewViewModel(
+    val repository: Repository,
+    val userId: String?,
```
Plouc314, line 23: Here you have the same problem as in #258, this should be `userid: String` or to be consistent with the navigation graph: `uid: String`

Baribal02 (approved):
Besides some details everything looks good

app/src/main/java/ch/epfl/skysync/components/forms/FlightForm.kt
```diff
@@ -504,6 +475,134 @@ fun RoleField(
...
+          Button(
+              modifier = Modifier.testTag("Add Role Button"),
+              onClick = {
+                roleNotChosenError = addNewRole == null
+                if (!roleNotChosenError) {
+                  onConfirm(addNewRole!!, addNewAssignee)
+                  addNewRole = null
+                  roleNotChosenError = false
+                  addNewAssignee = null
+                }
```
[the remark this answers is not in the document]
RAFInery: since I have a remember variable it has to be reset else it might show the last choices on the next alter

PR #289, opened 2024-05-05
User role distinction

Zep3k:
Created two graphs one for the admin and one for the crew and the pilot, in order to differentiate the screens between the admin and the crew/pilot without needing to check the type of the user everytime


app/src/main/java/ch/epfl/skysync/screens/LoadingScreen.kt
```diff
@@ -0,0 +1,24 @@
...
+import androidx.compose.runtime.getValue
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.navigation.NavHostController
+import ch.epfl.skysync.components.LoadingComponent
+import ch.epfl.skysync.models.user.Admin
+import ch.epfl.skysync.navigation.Route
+import ch.epfl.skysync.viewmodel.FlightsViewModel
+
+@Composable
+fun LoadingScreen(navController: NavHostController, viewModel: FlightsViewModel) {
```
Plouc314, line 13: I'm not sure to understand where this screen is used and what is its purpose ?

app/src/main/java/ch/epfl/skysync/screens/admin/AdminCalendar.kt
```diff
@@ -0,0 +1,92 @@
...
+import androidx.navigation.compose.currentBackStackEntryAsState
+import ch.epfl.skysync.components.AvailabilityCalendar
+import ch.epfl.skysync.components.FlightCalendar
+import ch.epfl.skysync.navigation.AdminBottomBar
+import ch.epfl.skysync.navigation.Route
+import ch.epfl.skysync.ui.theme.lightGray
+import ch.epfl.skysync.viewmodel.CalendarViewModel
+
+@Composable
+fun AdminCalendarScreen(
```
Plouc314, line 27: I don't know exactly the differences between the admin and crew/pilot version of the calendar screen,
* What are the UI difference, are there easy to modularize ? (My point is whether or not we can make a single calendar component and just pass different parameters to it)
* The view models will be different, but is the front end logic the same ? (That is, do we take the same callbacks ?)

app/src/main/java/ch/epfl/skysync/screens/admin/AdminChat.kt
```diff
@@ -0,0 +1,40 @@
...
+import androidx.compose.ui.platform.testTag
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.navigation.NavHostController
+import ch.epfl.skysync.components.GroupChat
+import ch.epfl.skysync.navigation.AdminBottomBar
+import ch.epfl.skysync.navigation.Route
+import ch.epfl.skysync.viewmodel.ChatViewModel
+
+@Composable
+fun AdminChatScreen(navController: NavHostController, viewModel: ChatViewModel) {
```
Plouc314, line 17: Is the admin chat any different from the crew/pilot one ? If not, I advise to keep a single chat/text screen directly in the `screens` folder (as it is done for the login screen for example)

app/src/main/java/ch/epfl/skysync/screens/crewpilot/Home.kt
```diff
@@ -0,0 +1,50 @@
...
+// Scaffold wrapper for the Home Screen
+@SuppressLint("UnusedMaterial3ScaffoldPaddingParameter")
+@Composable
+fun HomeScreen(navController: NavHostController, viewModel: FlightsViewModel) {
+  val currentFlights by viewModel.currentFlights.collectAsStateWithLifecycle()
+  Scaffold(
+      modifier = Modifier.fillMaxSize(),
+      bottomBar = { BottomBar(navController) },
+      floatingActionButton = {
+        // Define the FloatingActionButton to create a flight
```
Plouc314, line 28: The comment is misleading if we don't want to create flight on the crew/pilot screen

Plouc314 (commented):
Looks like it goes in the right direction, I'm just not up to speed with the details of each screen.

app/src/main/java/ch/epfl/skysync/screens/admin/AdminChat.kt
Zep3k: they are the same. I did two different screens in order to not have to check if the user is an admin or not every time

app/src/main/java/ch/epfl/skysync/screens/admin/AdminFlightDetailsScreen.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.navigation.NavHostController
+import ch.epfl.skysync.components.ConfirmFlightDetailUi
+import ch.epfl.skysync.components.FlightDetailUi
+import ch.epfl.skysync.models.flight.ConfirmedFlight
+import ch.epfl.skysync.navigation.AdminBottomBar
+import ch.epfl.skysync.navigation.Route
+import ch.epfl.skysync.viewmodel.FlightsViewModel
+
+@Composable
+fun AdminFlightDetailScreen(
```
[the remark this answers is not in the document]
Zep3k: The buttons on the bottom of the screen should not be there for the crew/pilot screen

app/src/main/java/ch/epfl/skysync/screens/LoadingScreen.kt
Zep3k: It is for when you connect to the app it takes some time before we have the information if the user is an admin or not. So while it's not yet decided the loading screen appears, then when it's decided it takes you to the right home screen.

app/src/main/java/ch/epfl/skysync/screens/admin/AdminChat.kt
Plouc314: Because the bottom bar is different, I see, then it makes sense to have two separate screens

app/src/main/java/ch/epfl/skysync/screens/LoadingScreen.kt
Plouc314: I get it, we have to refactore `onSignInResult` anyway (I'll try to do it in #263), that's good in the mean time

RAFInery (approved):

PR #290, opened 2024-05-05
user detail screen

mathieumognol:
#Close268 
Capture d'écran 2024-05-05 222016

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
+++ b/app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
@@ -108,6 +108,7 @@ fun NavGraphBuilder.homeGraph(
           ModifyFlightScreen(navController, flightsViewModel, flightId)
         }
     composable(Route.ADD_USER) { AddUserScreen(navController = navController) }
+
     composable(Route.USER) {
       val users =
           listOf(
--- a/app/src/main/java/ch/epfl/skysync/screens/UserDetails.kt
+++ b/app/src/main/java/ch/epfl/skysync/screens/UserDetails.kt
@@ -1,0 +1,283 @@
+package ch.epfl.skysync.screens
+
+import android.annotation.SuppressLint
+import android.location.Location
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxSize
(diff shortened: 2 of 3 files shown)


app/src/main/java/ch/epfl/skysync/components/FlightCard.kt
RAFInery, line 1: I think we could have some more hierarchy in the components (and screens). Maybe put it in a package:
components/userDetails ?

app/src/main/java/ch/epfl/skysync/screens/UserDetails.kt
```diff
@@ -0,0 +1,75 @@
...
+  val userFlights =
+      allFlights.filter { flight -> flight.team.roles.any { role -> role.assignedUser == user } }
+
+  Scaffold(
+      modifier = Modifier.fillMaxSize(),
+      topBar = { UserDetailTopAppBar(user = user, onBackClicked = { navController.navigateUp() }) },
+      bottomBar = { AdminBottomBar(navController = navController) },
+      floatingActionButtonPosition = FabPosition.End,
+  ) { paddingValues ->
+    FlightsList(userFlights, lightOrange, paddingValues, "Completed Flights") { selectedFlight ->
```
RAFInery, line 47: maybe take a different color for the title banner to better distinguish the home screen (upcoming flights) from the User details screen

app/src/main/java/ch/epfl/skysync/components/FlightsList.kt
```diff
@@ -0,0 +1,69 @@
...
+                    end = 16.dp,
+                    bottom = 16.dp),
+        color = Color.White,
+        textAlign = TextAlign.Center)
+
+    Spacer(modifier = Modifier.height(16.dp))
+    if (flights == null) {
+      LoadingComponent(isLoading = true, onRefresh = { /*TODO*/}) {}
+    } else if (flights.isEmpty()) {
+      // Handle case when no upcoming flights
```
RAFInery, line 55: left over comment from when this code was only used for the upcoming flights

app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
```diff
@@ -99,6 +99,15 @@ fun NavGraphBuilder.homeGraph(
           confirmationScreen(navController, flightId, flightsViewModel)
         }
 
+    /*composable(
```
RAFInery, line 102: I think you can already connect

RAFInery (approved):
LGTM

app/src/main/java/ch/epfl/skysync/navigation/HomeGraph.kt
mathieumognol: I think we need to create a userDetailsViewModel before that will give us a list of all the finished flights and fetch a user given a userId

app/src/main/java/ch/epfl/skysync/screens/UserDetails.kt
mathieumognol: Yes we could but I thought that it was nice to keep color coherence through the app especially for screen that are so far away

PR #291, opened 2024-05-06
Reports data model

Baribal02:
The data models for the report of the crew and the pilot

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/models/reports/CrewReport.kt
+++ b/app/src/main/java/ch/epfl/skysync/models/reports/CrewReport.kt
@@ -13,8 +13,7 @@ data class CrewReport(
     val prestigeChampagne: Int,
     override val begin: Date,
     override val end: Date,
-    override val pause: Boolean,
-    override val pauseDuration: Long, // in milliseconds
+    override val pauseDuration: Long?, // in milliseconds
     override val comments: String,
     override val vehicleProblems: Map<Vehicle, String>
 ) : Report
--- a/app/src/main/java/ch/epfl/skysync/models/reports/CrewReport.kt
+++ b/app/src/main/java/ch/epfl/skysync/models/reports/CrewReport.kt
@@ -2,6 +2,7 @@ package ch.epfl.skysync.models.reports
 
 import ch.epfl.skysync.models.flight.Vehicle
 import ch.epfl.skysync.models.user.Crew
+import java.sql.Time
 
 data class CrewReport(
     val id: String,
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/epfl/skysync/models/reports/Report.kt
```diff
@@ -0,0 +1,13 @@
+package ch.epfl.skysync.models.reports
+
+import ch.epfl.skysync.models.flight.Vehicle
+import java.sql.Time
+
+interface Report {
+  val begin: Time
```
Plouc314, line 7: Here you are using the `Time` class which is part of the sql package, instead consider using the `java.util.Date` class which is also used in the rest of the application

app/src/main/java/ch/epfl/skysync/models/reports/Report.kt
```diff
@@ -0,0 +1,13 @@
...
+  val end: Time
+  val pause: Boolean
+  val pauseDuration: Long
```
Plouc314, line 10: You didn't specify the unit of the pause duration (I'm assuming seconds but it would be neat to specify it explicitly).
Also, maybe we could have a single field `pause: Long?` that would be null when no pause have been taken ?

app/src/main/java/ch/epfl/skysync/models/reports/CrewReport.kt
```diff
@@ -0,0 +1,22 @@
...
+
+data class CrewReport(
+    val id: String,
+    val author: Crew,
+    val littleChampagne: Int,
+    val bigChampagne: Int,
+    val prestigeChampagne: Int,
+    val littleChampagneToFarmer: Boolean,
+    val bigChampagneToFarmer: Boolean,
+    val prestigeChampagneToFarmer: Boolean,
```
Plouc314, line 15: Isn't having three different fields to specify the type of champagne given to the farmer way too specific ?
Wouldn't this be a good use case of the `comments` fields ?

app/src/main/java/ch/epfl/skysync/models/reports/PilotReport.kt
```diff
@@ -0,0 +1,22 @@
+package ch.epfl.skysync.models.reports
+
+import ch.epfl.skysync.models.flight.Vehicle
+import ch.epfl.skysync.models.location.Location
+import ch.epfl.skysync.models.user.Pilot
+import java.sql.Time
+
+data class PilotReport(
+    val id: String,
```
Plouc314, line 9: Here you can pass a default value `val id: String = UNSET_ID` so that we can create an instance without having to specify an ID

app/src/main/java/ch/epfl/skysync/models/reports/Report.kt
Baribal02: It is in milliseconds as in other places where we use time as long but I forgot to precise it. I've done it now.

app/src/main/java/ch/epfl/skysync/models/reports/Report.kt
Baribal02: You're right it is easier to reuse it elsewhere

app/src/main/java/ch/epfl/skysync/models/reports/PilotReport.kt
Baribal02: Done

app/src/main/java/ch/epfl/skysync/models/reports/CrewReport.kt
Baribal02: Yeah, I've removed it.

Plouc314 (approved):
LGTM

PR #293, opened 2024-05-06
Bugfix persist modify vehicles

### LINKED ISSUE #279: bugfix: assign vehicles persistently
choice of vehicle in the modify screen does not get persisted to DB

RAFInery:
closes #279 

# problem
Setting a vehicle was not persisted to the DB.

# reason 
the list of vehicles was not properly saved on confirming the modifications

# solution
on save the correct list of vehicles is saved.

# other changes

- the list of vehicles is now immutable to prevent inexpected behavoiur


app/src/main/java/ch/epfl/skysync/components/forms/FlightForm.kt
```diff
@@ -98,11 +98,15 @@ fun FlightForm(
...
-        val listVehiclesValue = remember {
-          mutableStateListOf(*currentFlight?.vehicles?.toTypedArray() ?: emptyArray())
+        val chosenVehicles: MutableState<List<Vehicle?>> = remember {
+          if (currentFlight?.vehicles == null) {
+            mutableStateOf(listOf(null))
+          } else if (currentFlight.vehicles.isNotEmpty()) {
+            mutableStateOf(currentFlight.vehicles)
+          } else {
+            mutableStateOf(listOf(null))
+          }
```
Baribal02, line 108: Is it necessary to have a list of "Vehicle?" instead of a list of "Vehicle" ? And you could also eventually merge conditions as two of them are doing exactly the same thing.

Baribal02 (approved):
Besides minor details it seems all good to me

PR #295, opened 2024-05-06
E2E Staff Flight Detail

### LINKED ISSUE #183: End-to-end test: as staff I want to get more detailed information about a flight displayed in the overview when I click on it so that I can get more details for a specific flight

Prolas:
End to end test for the user story:
As staff I want to get more detailed information about a flight displayed in the overview when I click on it, so that I can get more details for a specific flight.

We should unify the UI of a planned and confirmed flight as they are similar.

close #183


mathieumognol (approved):
LGTM but you could've used helper methods to avoid repetition when verifying the display of properties like balloon names, basket names, etc

PR #297, opened 2024-05-06
Cancel availability calendar

### LINKED ISSUE #282: Cancel function calendar viewmodel

Prolas:
Added cancel function in  CalendarViewModel.
When clicking the cancel button in AvailabilityCalendar the modified availabilities disapear.

close #282


app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
```diff
@@ -159,4 +159,23 @@ class CalendarViewModel(
         // However this might be unnecessary
         refreshUser()
       }
+
+  /** Cancels the modification made in the current availability calendar */
+  fun cancelAvailabilities() =
```
RAFInery, line 164: couldn't we just reset the availabilities of the user with `originalAvailabilityCalendar` that contains the persitent state of the AvailabilityCalendar before the modifications?

RAFInery (approved):

app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
Prolas: If you're thinking of `user.availabilities = originalAvailabilityCalendar`, this isn't possible because user.availabilities is immutable. But if you're thinking of something else, I'd like to know.

RAFInery: why is that a problem?

Prolas: User.availabilities is val and can not be reassigned.

RAFInery: and making just a copy of the user with the new availabilities?

PR #300, opened 2024-05-08
modify/add flight uses filters for available equipment and users

### LINKED ISSUE #243: connect new filtered DB functions to FlightsViewModel
- inject the functions from #219 into the `FlightsViewModel`
- update tests

RAFInery:
closes #243 

# connected availability filters

- for a given date & timeSlot only the equipment that has not been assigned to an other flight on the same date&timeSlot can be chosen
- for a given date & timeSlot only the users that have not been assigned to an other flight on the same date&timeSlot and that are available can be chosen

# added back the ability to delete a role
- the delete button was temporarily invisible so it was made visible again


app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
```diff
@@ -73,6 +75,14 @@ class FlightsViewModel(
...
   val availableUsers = _availableUsers.asStateFlow()
 
+  //    private var currentFlight: StateFlow<Flight?> =
+  //        combine(currentFlightId, _currentFlights) { fid, flights ->
+  //            flights?.find { it.id == (fid ?: "") }
+  //        }.stateIn(
+  //                scope = viewModelScope,
+  //                started = WhileUiSubscribed,
+  //                initialValue = null)
+
```
Zep3k, line 85: why is this commented?

Zep3k (approved):
LGTM

PR #303, opened 2024-05-09
User global view model

### LINKED ISSUE #263: Global view model for user

Plouc314:
* Create `UserGlobalViewModel` that fetch the user (and handle temp user / default user) after the sign in
* Update `MainGraph` to use this view model
* Did not update `homeGraph` and further down to take a `User` object instead of `uid` is this would require to refactore the whole navigation graph, the view models and the tests. Also, with the #289 PR open in parallel this would create a lot of complications

Close #263


Prolas:
LGTM

Prolas (approved):
LGTM

PR #304, opened 2024-05-09
Google map links

### LINKED ISSUE #271: Explore link to google maps
* For itinerary of crew to landing point
* For meeting point with passenger before flight

Zep3k:
Now when you add a location to a confirmed flight it creates a hyperlink such that when you click on the location it takes you to google maps
Screenshot_20240509-123252_SkySync
Screenshot_20240509-123254_Maps

Close #271


app/src/main/java/ch/epfl/skysync/components/ConfirmFlightDetailUI.kt
```diff
@@ -119,6 +125,41 @@ fun ListElementText(text: String, padding: Dp, testTag: String) {
   )
   Spacer(modifier = Modifier.padding(4.dp))
 }
+
+@Composable
+fun HyperLinkText(title: String, value: String, padding: Dp) {
```
mathieumognol, line 130: Maybe calling it `location` rather than `value` would make more sense in this context

app/src/main/java/ch/epfl/skysync/components/ConfirmFlightDetailUI.kt
```diff
@@ -119,6 +125,41 @@ fun ListElementText(text: String, padding: Dp, testTag: String) {
...
+  val uriHandler = LocalUriHandler.current
+  val googleMapsLink = "https://www.google.com/maps/search/?api=1&query=${value.replace(" ", "+")}"
```
mathieumognol, line 132: Could be more careful to do:
val encodedValue = URLEncoder.encode(value, "UTF-8")
val googleMapsLink = "https://www.google.com/maps/search/?api=1&query=$encodedValue"
In case there are other special characters in the value string

app/src/main/java/ch/epfl/skysync/components/ConfirmFlightDetailUI.kt
```diff
@@ -119,6 +125,41 @@ fun ListElementText(text: String, padding: Dp, testTag: String) {
...
+
+  ClickableText(
+      text = string,
+      modifier = Modifier.padding(horizontal = padding.plus(4.dp)).testTag(title),
+      style = MaterialTheme.typography.bodyLarge,
+      onClick = { offset ->
+        string.getStringAnnotations(tag = "URL", start = offset, end = offset).firstOrNull()?.let {
+            stringAnnotation ->
+          uriHandler.openUri(stringAnnotation.item)
+        }
```
mathieumognol, line 159: Could add error handling:
?: run {
                // Handle the case where no URL is found
            }

mathieumognol (approved):
LGTM

app/src/main/java/ch/epfl/skysync/components/ConfirmFlightDetailUI.kt
Zep3k: I think there will always be an url because i created it on line 132:
val googleMapsLink = "https://www.google.com/maps/search/?api=1&query=${value.replace(" ", "+")}"
then i build a string with tag url with it

app/src/main/java/ch/epfl/skysync/components/ConfirmFlightDetailUI.kt
Zep3k: thank you

PR #306, opened 2024-05-09
combined locationViewModel and timerViewModel

### LINKED ISSUE #264: Shared view model for flight
* Superset of timer view model
* Manage flight: timer, location sharing
* Create finished flight (redirect somewhere ?)

RAFInery:
closes #264 

# combined `locationViewModel` and `timerViewModel` into one VM
The new ViewModel now has the two major functions
- `startFlight`: to start the counter and trace capturing
- `stopFlight`: to stop the counter and the trace capturing. Also a function `saveFinishedFlight` is called that for the moment does nothing but later on will be in charge of converting  the ConfirmedFlight into a FinishedFlight and saving it to the DB

# scetch for cockpit view
To choose the flight that wants to be started a blueprint screen was added that lets the user pick a flight 

# comments
- before calling the start or stop function, be sure that the userId was set and `refreshPersonalFlights` called at least once
- as the timerViewModel before, the locationViewModel is defined on the toplevel `MainActivity` to assure existence over all screens. To switch to a shared view model a new constructor must be defined


app/src/main/java/ch/epfl/skysync/navigation/MainGraph.kt
```diff
@@ -17,12 +17,12 @@ fun MainGraph(
     navHostController: NavHostController,
     signInLauncher: ActivityResultLauncher<Intent>,
     uid: String?,
-    timer: TimerViewModel
+    inFlightsViewModel: LocationViewModel? = null
```
Plouc314, line 20: Why is `inFlightsViewModel` marked as nullable ?

app/src/main/java/ch/epfl/skysync/viewmodel/LocationViewModel.kt
```diff
@@ -8,23 +9,30 @@ import androidx.lifecycle.viewmodel.compose.viewModel
...
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
 import kotlinx.coroutines.launch
 
-class LocationViewModel(val uid: String, repository: Repository) : ViewModel() {
+/** ViewModel for the location tracking of the user during a flight and the timer. */
+class LocationViewModel(uid: String? = null, val repository: Repository) : ViewModel() {
```
Plouc314, line 31: Maybe rename the view model to something clearer as it does not handle only locations now

Plouc314 (approved):
LGTM, just small details

app/src/main/java/ch/epfl/skysync/navigation/MainGraph.kt
RAFInery: for testing purposes and to reduce refactoring of present tests

app/src/main/java/ch/epfl/skysync/viewmodel/LocationViewModel.kt
RAFInery: I agree, but we'll do that in a separate PR since renaming classes combined with a lot of modifications does confuse  Git and I wanted that changes are clearly visible

Issue #315, opened 2024-05-10
Confirm flight details


Plouc314:
* Remove id
* location is hardcoded

PR #329, opened 2024-05-12
Refactor ConfirmFlightScreen 

### LINKED ISSUE #313: Confirm flight UI refactor
* departure time/meetup: time picker
* Passenger meetup location: not clear
* add back button
* if remark too long it covers to button
* change colors

Prolas:
* Refactor ConfirmFlightScreen 
 * Add custom text in CustomText
 * Creation of generic Time Picker
 * Add TimePickerButton: button which opens a time picker when clicking on it
 * Refactor date utility funtcions from FormUtils to DateUtility
 * Change in FlightColor YELLOW -> GREEN
 close #313  #315 
# Admin flight details
The only difference between this screen and the crew is that it doesn't have the buttons.
## Planned flight
image

## Confirmed flight
image
image

 # Admin vue to confirm a flight
image
image


app/src/main/java/ch/epfl/skysync/components/CustomTexts.kt
Baribal02, line 1: Might be useful to also put the modifier as a parameter. You could also put defalut values for values that does not change usually like colors or potentially modifier.

app/src/main/java/ch/epfl/skysync/database/DateUtility.kt
```diff
@@ -57,4 +67,40 @@ object DateUtility {
...
+   *
+   * @param date The [LocalDate] object to be converted.
+   * @return The equivalent [LocalTime] object or --/--/-- if null.
+   */
+  fun localDateToString(date: LocalDate?): String {
+    return date?.format(DateTimeFormatter.ofPattern("dd/MM/yy")) ?: "--/--/--"
+  }
+
+  /**
+   * Retrieves the hours and minutes from a [Date] object and returns a String. I null returns --:--
```
Baribal02, line 94: Typo in the comment. It should be "If null returns ..."

app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
```diff
@@ -91,6 +91,7 @@ class FirestoreDatabase(private val useEmulator: Boolean = false) {
   ): T? {
     val documentSnapshot = db.collection(path).document(id).get().await()
 
+    println(clazz.java)
```
Baribal02, line 94: You shouldn't have a print there. If you want to show some informations you should use the Log class

app/src/main/java/ch/epfl/skysync/screens/admin/AdminFlightDetailsScreen.kt
```diff
@@ -23,38 +50,123 @@ fun AdminFlightDetailScreen(
...
-          confirmClick = { navController.navigate(Route.CONFIRM_FLIGHT + "/${flightId}") },
-          padding = padding,
-          flight = flight,
-          bottom = { editClick, confirmClick, deleteClick,
-            ->
-            FlightDetailBottom(editClick, confirmClick, deleteClick)
-          })
-    }
+  val showConfirmDialog = remember { mutableStateOf(false) }
+  if (showConfirmDialog.value) {
```
Baribal02, line 54: Nitpick : you could use `val showConfirmDialog by remember { mutableStateOf(false) }` to avoid all the .value later in the code

Baribal02 (approved):
LGTM. These changes make the code overall clearer

app/src/main/java/ch/epfl/skysync/components/CustomTexts.kt
Prolas: I added the modifier as parameter with default value `Modifier` and a Color.Black as default value for the tiltes

app/src/main/java/ch/epfl/skysync/database/DateUtility.kt
Prolas: fixed

app/src/main/java/ch/epfl/skysync/database/FirestoreDatabase.kt
Prolas: Fixed. This was for testing purposes and we do not need a log here.

app/src/main/java/ch/epfl/skysync/screens/admin/AdminFlightDetailsScreen.kt
Prolas: I will do that

PR #334, opened 2024-05-13
Ongoing flight backend

### LINKED ISSUE #321: Ongoing flight DB/logic
* Store in the dabatase the ongoing flights (ex: timer)
* Logic to first query the ongoing flights when going on flight screen
* Only the pilot can start a flight

### LINKED ISSUE #322: Ongoing flight setup listener on start/stop flight
Setup listener to notify the flight members that the flight is started/finished

Plouc314:
* Rename `LocationViewModel` to `InFlightViewModel`
* `InFlightViewModel` has 3 stages `IDLE`, `ONGOING`, `POST`, the `POST` stage is new, it is when the flight is finished (when the balloon touch the ground) but the user want to stay on the flight screen (to see the balloon on the map), in this stage everything is static. There is now 3 buttons `Start Flight`, `Stop Flight` and `Clear Flight` which switch from `POST` to `IDLE`, clear all data about the current flight.
* Only the pilot can start, stop, clear a flight
* Add a listener to be notified when a flight is started/stopped

Close #321
Close #322

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/navigation/CrewPilotGraph.kt
+++ b/app/src/main/java/ch/epfl/skysync/navigation/CrewPilotGraph.kt
@@ -33,7 +33,7 @@ import ch.epfl.skysync.screens.reports.CrewReportScreen
 import ch.epfl.skysync.screens.reports.PilotReportScreen
 import ch.epfl.skysync.viewmodel.ChatViewModel
 import ch.epfl.skysync.viewmodel.FlightsViewModel
-import ch.epfl.skysync.viewmodel.LocationViewModel
+import ch.epfl.skysync.viewmodel.InFlightViewModel
 import ch.epfl.skysync.viewmodel.MessageListenerSharedViewModel
 import java.time.LocalDate
 import java.util.Date
@@ -42,7 +42,7 @@ fun NavGraphBuilder.crewPilotGraph(
     repository: Repository,
     navController: NavHostController,
     uid: String?,
-    locationViewModel: LocationViewModel? = null
+    inFlightViewModel: InFlightViewModel? = null
 ) {
   navigation(startDestination = Route.CREW_HOME, route = Route.CREW_PILOT) {
     personalCalendar(repository, navController, uid)
@@ -74,6 +74,10 @@ fun NavGraphBuilder.crewPilotGraph(
         }
     composable(Route.CREW_HOME) { entry ->
 
+      // initiate in flight view model here, so that we can notify
+      // the user when a flight is started by someone else
+      inFlightViewModel!!.init(uid!!)
+
       // get the MessageListenerSharedViewModel here so that it gets
       // instantiated
       val messageListenerSharedViewModel =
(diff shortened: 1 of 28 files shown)


app/src/main/java/ch/epfl/skysync/navigation/CrewPilotGraph.kt
```diff
@@ -74,6 +74,10 @@ fun NavGraphBuilder.crewPilotGraph(
         }
     composable(Route.CREW_HOME) { entry ->
 
+      // initiate in flight view model here, so that we can notify
+      // the user when a flight is started by someone else
+      inFlightViewModel!!.init(uid!!)
```
RAFInery, line 79: again?

app/src/main/java/ch/epfl/skysync/viewmodel/InFlightViewModel.kt
```diff
@@ -0,0 +1,506 @@
...
+
+  /** If the current flight stage is [FlightStage.POST] */
+  fun isPostFlight(): Boolean = _flightStage.value == FlightStage.POST
+
+  /** Returns if the user is the pilot of the ongoing flight */
+  fun isPilot(): Boolean = pilotId == _userId
+
+  /** Returns if the user is assigned a pilot role in the team */
+  private fun isUserPilotRole(team: Team): Boolean {
+    return team.roles.find { role ->
```
RAFInery, line 173: maybe define in `Team.kt`?

app/src/main/java/ch/epfl/skysync/viewmodel/InFlightViewModel.kt
```diff
@@ -0,0 +1,506 @@
...
+  private val _flightLocations = MutableStateFlow<List<Location>>(emptyList())
+
+  /** List of the successive locations through which the balloon (pilot) has passed */
+  val flightLocations: StateFlow<List<Location>> = _flightLocations.asStateFlow()
+
+  /** List of confirmed flights scheduled today where the user is assigned */
+  private val _confirmedFlights: MutableStateFlow<List<ConfirmedFlight>> =
+      MutableStateFlow(emptyList())
+
+  val startableFlights: StateFlow<List<ConfirmedFlight>> =
```
RAFInery, line 135: Each day a user can be assigned to max 2 flights. Instead of showing both flights maybe we want to show only the current one i.e. in the morning we show the morning flight and in the evening the evening flight (if the user is assigned to a flight)

app/src/main/java/ch/epfl/skysync/screens/crewpilot/Flight.kt
```diff
@@ -181,20 +183,21 @@ fun FlightScreen(
...
-  } else if (personalFlights == null) {
+  } else if (loading) {
     LoadingComponent(isLoading = true, onRefresh = {}) {}
-  } else if (personalFlights!!.isEmpty()) {
-    ShowFlightToStart(navController = navController, flight = null) {}
-  } else if (currentFlightId == null) {
-    ShowFlightToStart(navController, personalFlights!!.first()) {
-      inFlightViewModel.setFlightId(it)
+  } else if (currentFlight == null) {
+    ShowFlightToStart(navController, startableFlights.firstOrNull()) {
```
RAFInery, line 199: if we have two flights (AM and PM) it might choose one of the two at random?

RAFInery (approved):
Nice work. The only question that remains is if we want to show:
a) only the current flight -> In the morning we show the AM flight (if any), in the evening the PM flight (if any)
b) both flights of a day (if any) but let the user only start the flight belonging to the current TimeSlot

Also, a flight that has been finished, can it be restarted or is it supposed to disappear from the "start flight screen"?

app/src/main/java/ch/epfl/skysync/viewmodel/InFlightViewModel.kt
Plouc314: Make sense, gone for it

app/src/main/java/ch/epfl/skysync/screens/crewpilot/Flight.kt
Plouc314: Doesn't happen anymore as we only consider flight in current time slot (either AM or PM)

Plouc314:
The flight will be updated to `FinishedFlight` once it's finished, so it should not be possible to restart it.

PR #336, opened 2024-05-14
Make calendar immutable

### LINKED ISSUE #311: Fix calendar related bugs
Known bugs:
* Save availability 2 times in a row crashes the application
* Cancel function does not update the UI 

Possible solutions:
* The UI react to reference changes, thus changing an immutable list doesn't trigger a recompose, refactor the CalendarModel to use an immutable list that would be updated (basically make CalendarModel an immutable class)

### LINKED ISSUE #324: refactor Calendar: Mutable to Immutable

RAFInery:
closes #324 

# `CalendarModel` is now immutable 
- changes result in a new CalendarModel instance that have the modifcations

# the calendars are now longer maintained in the `User` object
- the calendar information and the user information are fetched separately from the DB. Furthermore the Calendar data is only used in the Calendar screen as opposed to the User data that is shared to multiple screens and viewmodels.
- The corresponding fields in the `User`  are thus no longer needed and were deleted

# bugfix: 
On changing to immutable the bug was resolved.
closes #311 

## problem

the cancel button did not work correctly => Canceling an added availability worked fine but modifications on the other hand were not reversed.
## reason
the `equals` functions of CalendarModel was badly implemented such that two Calendar instances were considered "equal" if they had the same amount of cells but possibly different values for the cells. This caused the calendar and its current changes that were saved in a MutableFlow in the CalendarViewModel to be ignored since a Flow is only updated when its value _changes_. 

## fix 
The equals overwrite function was removed. Since the Calendar is now immutable, two Calendars can be compared by their reference


app/src/main/java/ch/epfl/skysync/screens/admin/AdminCalendar.kt
```diff
@@ -43,21 +43,21 @@ fun AdminCalendarScreen(
...
               padding = padding,
               getAvailabilityStatus = { date, time ->
                 availabilityCalendar.getAvailabilityStatus(date, time)
               },
               nextAvailabilityStatus = { date, time ->
-                availabilityCalendar.nextAvailabilityStatus(date, time)
+                viewModel.setToNextAvailabilityStatus(date, time)
               },
               onSave = { viewModel.saveAvailabilities() },
               onCancel = { Log.d("TO BE IMPLEMENTED", "Cancel in Availabilities") })
```
Prolas, line 58: I think the cancel function is missing here

app/src/androidTest/java/ch/epfl/skysync/viewmodel/CalendarViewModelTest.kt
```diff
@@ -33,105 +31,124 @@ class CalendarViewModelTest {
...
-            dbs.availability1Admin1.date, dbs.availability1Admin1.timeSlot))
+            dbs.availability1Crew1.date, dbs.availability1Crew1.timeSlot))
     assertEquals(
-        AvailabilityStatus.OK,
+        AvailabilityStatus.UNDEFINED,
+        availabilityCalendar.getAvailabilityStatus(LocalDate.of(1999, 10, 1), TimeSlot.AM))
+  }
+
+  @Test
+  fun testSetToNextAvailabilityStatusForExisting() = runTest {
```
Prolas, line 52: Would it be possible to use a for loop here ?
```
val availabilityStatus = listOf(AvailabilityStatus.OK, AvailabilityStatus.MAYBE, AvailabilityStatus.NO, AvailabilityStatus.UNDEFINED)
availabilityStatus.foreach {}
```

app/src/main/java/ch/epfl/skysync/components/AvailabilityCalendar.kt
```diff
@@ -154,12 +154,13 @@ fun AvailabilityCalendar(
       // it each time the result of getAvailabilityStatus change
       // which is a bit hacky and should be a temporary solution
       val availabilityStatus = getAvailabilityStatus(date, time)
-      var status by remember { mutableStateOf(availabilityStatus) }
-      if (status != availabilityStatus) {
-        status = availabilityStatus
-      }
-      AvailabilityTile(date = date, time = time, availabilityStatus = status) {
-        status = nextAvailabilityStatus(date, time)
+      // var status by remember { mutableStateOf(availabilityStatus) }
```
Prolas, line 157: If not used anymore this should be deleted

Prolas (commented):
Did you experience any crashes ? 
My emulator crashes in this scenario:
* Save an availability 
* Click on it until it has the same color (click 4 times) and save the availabiltiy

app/src/main/java/ch/epfl/skysync/models/calendar/CalendarModel.kt
```diff
@@ -48,24 +44,31 @@ abstract class CalendarModel<T : CalendarViewable>(
...
    * @param timeSlot the timeSlot of the cell to remove
    * @return the removed value if it was found, null otherwise
    */
-  protected fun removeByDate(date: LocalDate, timeSlot: TimeSlot): T? {
+  protected fun removeByDate(date: LocalDate, timeSlot: TimeSlot): CalendarModel<T> {
 
     val oldValue = getByDate(date, timeSlot)
-    if (oldValue != null) {
-      cells.remove(oldValue)
+    return if (oldValue != null) {
```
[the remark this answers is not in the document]
RAFInery: it's been suggested to me by the IDE

RAFInery:
You were right. I just had to change the comparison mentioned by you. Now the bug should be gone

Prolas (approved):
LGTM

PR #342, opened 2024-05-15
availability calendar

hacknstack:
Fixed availability crash when you changed your availabilities when a flight was already assigned.
- There's now a new color for assigned flight which makes the UI more similar the Figma and more understandable to use.


Zep3k (approved):
LGTM

PR #343, opened 2024-05-15
Error handling filter exception

Zep3k:
the job was cancelled is now filtered out


hacknstack (approved):
LGTM, glad to not have the job canceled message no more 😄

PR #345, opened 2024-05-15
UI fixing

### LINKED ISSUE #339: Fix UI
Fix UI colors padding and utilities to make it more uniform practical and intuitive

hacknstack:
Description
------------------------

### Overview
Close #339

Rework of UI

### Changes Made

*   Redone login for it to look like the figma prototype
*   Added text describing states in availability calendar 
*   Reworked the delete,edit,confirm button 
*  Added necessary padding for all the flight in flightlist to show
*  Reworked user screen to look like the figma prototype
### Android renders:
calendarfordaltonian
login
User screen
zdeletemodconfirm


app/src/main/java/ch/epfl/skysync/components/FlightsList.kt
```diff
@@ -63,7 +63,7 @@ fun FlightsList(
           }
     } else {
       // Display the flights in a LazyColumn if the list is not empty
-      LazyColumn(modifier = Modifier.testTag("HomeLazyList")) {
+      LazyColumn(modifier = Modifier.testTag("HomeLazyList").padding(bottom = 64.dp)) {
```
Baribal02, line 66: You should put the padding on the column and you should use the paddingValues that is given as parameter. It is the padding of the Scaffold.

app/src/main/java/ch/epfl/skysync/screens/admin/AdminFlightDetailsScreen.kt
```diff
@@ -102,26 +110,60 @@ fun FlightDetailBottom(
...
-          ClickButton(
-              text = "Confirm",
-              onClick = { confirmClick() },
-              modifier = Modifier.fillMaxWidth(0.7f).testTag("ConfirmButton"),
-              color = lightGreen)
-        }
+    Row(modifier = Modifier.fillMaxWidth(), verticalAlignment = Alignment.CenterVertically) {
+      Spacer(modifier = Modifier.fillMaxWidth(0.05f))
+      Row(modifier = Modifier.fillMaxWidth(0.9f / 3 / (1 - 0.05f)).testTag("DeleteButton")) {
+        TextButton(
```
Baribal02, line 116: You don't need to make such complicated calculation for the fillMaxWidth. You can use the weights. For example to have each rom taking the same space you could have `Modifier.fillMaxWidth.weight(1f)` for each row that you want of the same size. Note that, for it to work you will also need to embed this rows in a Row or a Column (which is the case here).

Baribal02 (approved):
Some minor things to change but overall looks good to me.

PR #349, opened 2024-05-16
calendar saving bug

### LINKED ISSUE #348: bugfix: Calendar crash
# problem

When saving an availability is changed by passing through `UNDEFINED` e.g. `NO` - > `UNDEFINED` -> `OK`
then the app crashes

RAFInery:
closes #348 

# solution

On changing an initially defined availability to UNDEFINED removed the availability from the Calendar. On changing further to a different status the ID is no longer known however when computing the differences in `getDifferencesWithOtherCalendar` the change is considered from a defined state to another defined state thus he will try to update the availability but has no id.

# fix
use `UNDEFINED` to mark deleted availabilities that have not yet been synchronized with the DB instead of deleting it. 
For the change `NO` -> `Ok` the backend now does:
`NO` -> `UNDEFINED` -> `OK`
instead of 
`NO` -> (removed) -> (newly init) `OK`


hacknstack (approved):
LGTM good fix 👍

PR #352, opened 2024-05-16
Chat issues

Prolas:
# Issue
* Chat screen: the chat was updated only when the page was loaded. So when a message arrived and we were on the screen, nothing happened.
* Other screens: when a message was received, the snackbar didn't show up 

# Bug

For messages, a shared viewModel has been used and initialized in the home screen. This lives on until the home screen is pop from the stack. However, when navigating with the bottomBar, all screens up to the login screen (excluded) are pop.

# Solution
I've created a viewModel similar to the inFlightViewModel, which is created in the login screen passed as an argument to the graph, then initialized in the home screen.


app/src/main/java/ch/epfl/skysync/viewmodel/MessageListenerViewModel.kt
```diff
@@ -46,17 +47,21 @@ class MessageListenerSharedViewModel : ViewModel() {
...
                 override fun <T : ViewModel> create(modelClass: Class<T>): T {
-                  return MessageListenerSharedViewModel() as T
+                  return MessageListenerViewModel() as T
                 }
               })
     }
   }
 
+  init {
+    println("Init shared view model")
```
Plouc314, line 62: You can remove the prints now

Plouc314 (approved):
LGTM

app/src/main/java/ch/epfl/skysync/viewmodel/MessageListenerViewModel.kt
Prolas: Indeed

PR #353, opened 2024-05-16
Finished flight view model

Baribal02:
This viewModel is able to : 
- add a finishedFlight
- get all the finishedFlights linked to the current user
- add a report to a given flight
- get all reports of a given flight
- select a flight


settings.gradle.kts
```diff
@@ -19,6 +19,6 @@ dependencyResolutionManagement {
     }
 }
 
-rootProject.name = "SkySync"
+rootProject.name = "skysync"
```
hacknstack, line 22: I don't see a reason to change this

hacknstack (changes requested):
LGTM overall, need only 1 fix imo

settings.gradle.kts
Baribal02: Done

hacknstack (approved):

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+import ch.epfl.skysync.util.WhileUiSubscribed
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.stateIn
+import kotlinx.coroutines.launch
+
+/** ViewModel for the user for the finished flights */
+class FinishedFlightsViewModel(val repository: Repository, val userId: String?) : ViewModel() {
```
Plouc314, line 27: You should take a non-nullable user id `val userId: String`, at this place in the application we know the uid, this will simplify the logic and remove unnecessary checks

app/src/main/java/ch/epfl/skysync/screens/reports/CrewReport.kt
```diff
@@ -20,123 +21,141 @@ import androidx.compose.ui.Modifier
...
+              errorPrestigeChampagne = !bottleInputValidation(prestigeChampagne)
+              if (!hasError(errorLittleChampagne, errorBigChampagne, errorPrestigeChampagne)) {
+                CrewReport(
+                    author = crew.value!!.id,
+                    begin = beginTime,
+                    end = endTime,
+                    pauseDuration = pauseDuration.toInt(),
+                    comments = comments,
+                )
+                // TODO save report
```
Plouc314, line 150: Should `finishedFlightsViewModel.addReport` be called here ?

app/src/main/java/ch/epfl/skysync/screens/reports/PilotReport.kt
```diff
@@ -21,134 +21,147 @@ import androidx.compose.ui.Modifier
...
+                    takeOffTime = takeoffTime,
+                    takeOffLocation = takeoffLocation,
+                    landingLocation = landingLocation,
+                    landingTime = landingTime,
+                    begin = beginTime,
+                    end = endTime,
+                    pauseDuration = pauseDuration.toInt(),
+                    comments = comments,
+                )
+                // TODO save report
```
Plouc314, line 156: Same here

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+        isLoading.value = false
+      }
+
+  fun addFlight(flight: FinishedFlight) =
+      viewModelScope.launch {
+        repository.flightTable.add(flight, onError = { onError(it) })
+        refreshUserAndFlights()
+      }
+
+  fun deleteFlight(flightId: String) =
```
Plouc314, line 95: From what I understand `FinishedFlightsViewModel` is used in the reports screen, when would you need to add/delete a flight there ? These functions already exists in the `FlightViewModel` which is used in the `AdminFlightDetailScreen`

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+
+  fun refreshAndSelectFlight(id: String) =
+      viewModelScope.launch {
+        refreshUserAndFlights().join()
+        selectFlight(id)
+      }
+
+  /** Callback executed when an error occurs on database-related operations */
+  private fun onError(e: Exception) {
+    SnackbarManager.showMessage(e.message ?: "An unknown error occurred")
```
Plouc314, line 133: Here you should add a check (see `FlightsViewModel.onError` for reference)

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -0,0 +1,135 @@
...
+  fun getAllReports(flightId: String) =
+      viewModelScope.launch {
+        _flightReports.value =
+            repository.reportTable.retrieveReports(flightId, onError = { onError(it) })
+        Log.d("FinishedFlightsViewModel", "Reports retrieved : ${_flightReports.value}")
+      }
+
+  private fun selectFlight(id: String) = viewModelScope.launch { _selectedFlightId.value = id }
+
+  fun refreshAndSelectFlight(id: String) =
```
Plouc314, line 125: This function is never used (except in the tests), and you need to call this function to set `_selectedFlightId`, so `_selectedFlightId` is never set, so `addReport` will never work. Maybe you could drop the `_selectedFlightId` attribute altogether and pass a `flightId` argument to `addReport`?

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
Baribal02: Changed it

app/src/main/java/ch/epfl/skysync/screens/reports/CrewReport.kt
Baribal02: Yes, you're right

app/src/main/java/ch/epfl/skysync/screens/reports/PilotReport.kt
Baribal02: Done

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
Baribal02: It's added

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
Baribal02: I have mixed up to ideas. I'll keep your suggestion as it will make the viewModel simpler

PR #379, opened 2024-05-18
fix spam back click button

Prolas:
## Bug
When using the “back” button, the associated function was sometimes multiple times, the `popBackStack()` returns false because there are no screen left on the stack leaving the user with a blank screen.

## Solution
When the `popBackStack()` returns false navigate to the last screen before the blank screen


app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
Baribal02, line 1: You could also change other hardcoded string for the ids

Baribal02 (approved):
LGTM

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
Prolas: I'll do it

PR #384, opened 2024-05-20
Fix inflight flight trace

### LINKED ISSUE #359: Fix flight trace bug
When quitting the app during a flight, and reopening the app, displayed flight trace starts from the current location of the balloon and not the one where it took of.

Plouc314:
* Fixed a bug that would provoke users that would go on the flight screen in the middle of the flight to only see the flight trace from where the balloon was when the user joined
* Added a test for the bug
* Fixed a bug in `FlightsViewModel` that would always fetch all balloon
* Fixed `FlightViewModelTest`, some concurrency issues and a test that was depended on the time of the day

Close #359


RAFInery (approved):
LGTM

PR #385, opened 2024-05-20
Crew/Pilot Stats Screen UI

### LINKED ISSUE #357: Crew pilot stat screen
* The user (crew/pilot) should be able to see his finished flights

mathieumognol:
Close #357

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/screens/crewpilot/Stats.kt
+++ b/app/src/main/java/ch/epfl/skysync/screens/crewpilot/Stats.kt
@@ -1,0 +1,35 @@
+package ch.epfl.skysync.screens.crewpilot
+
+import android.util.Log
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material3.FabPosition
+import androidx.compose.material3.Scaffold
+import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


app/src/main/java/ch/epfl/skysync/navigation/BottomNavigationBar.kt
```diff
@@ -14,7 +14,7 @@ fun BottomBar(navController: NavHostController) {
           BottomBarScreen.Flight,
           BottomBarScreen.Chat,
           BottomBarScreen.Calendar,
-      )
+          BottomBarScreen.Stats)
```
Zep3k, line 17: You should also modify AdminBottomNavigationBar. Because like this when you click on the stats while being an admin it takes you to the crew/pilot Stat screen

Zep3k (approved):
Except what i commented it looks good. It would also be nice to have screenshots to make it easier to review the Ui

app/src/main/java/ch/epfl/skysync/navigation/BottomNavigationBar.kt
mathieumognol: Fixed

PR #387, opened 2024-05-20
Update dateToHourMinuteString to use right timeZone

### LINKED ISSUE #362: Time picker change time zone

Prolas:
# Bug
The `dateToHourMinuteString` function converted a date with the wrong time zone, which led to the display of an incorrect time but did not affect the time stored in the database. This function was used in ConfirmFlightScreen.

# Solution
Replace the time zone in use with the time zone of the device in use.

close #362


RAFInery (approved):
LGTM

PR #390, opened 2024-05-21
Time range picker

Baribal02:
This is a component that allows a user to pick a time duration
It is for example used for reports form to get the total pause time.

image

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/components/RangeTimePicker.kt
+++ b/app/src/main/java/ch/epfl/skysync/components/RangeTimePicker.kt
@@ -1,0 +1,144 @@
+package ch.epfl.skysync.components
+
+import androidx.compose.animation.animateColorAsState
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.foundation.lazy.LazyListState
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/epfl/skysync/components/RangeTimePicker.kt
```diff
@@ -0,0 +1,147 @@
...
+    title: String,
+    showDialog: Boolean,
+    onDismiss: () -> Unit,
+    onConfirm: (Long) -> Unit
+) {
+  if (!showDialog) {
+    return
+  }
+  var timeDuration by remember { mutableLongStateOf(0L) }
+  val listState1 = rememberLazyListState(Int.MAX_VALUE / 2)
```
mathieumognol, line 43: Could use more descriptive cariable name like for example: `hourListState`

app/src/main/java/ch/epfl/skysync/components/RangeTimePicker.kt
```diff
@@ -0,0 +1,147 @@
...
+  val listState2 = rememberLazyListState(Int.MAX_VALUE / 2)
```
mathieumognol, line 44: Could use more descriptive cariable name like for example: `minuteListState`

mathieumognol (approved):
LGTM

app/src/main/java/ch/epfl/skysync/components/RangeTimePicker.kt
Baribal02: Done

PR #394, opened 2024-05-22
UI enhancement

Zep3k:
- Unified the headers of the main screens
- fixed shape of buttons on flight detail
- fixed sign in button text, it took two lines depending on the device
- added padding to user management screen
- fixed ui of user details
Screenshot_20240522-162552_SkySync
Screenshot_20240522-162557_SkySync
Screenshot_20240522-162403_SkySync
Screenshot_20240522-163716_SkySync
Screenshot_20240522-162606_SkySync

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/components/GroupChat.kt
+++ b/app/src/main/java/ch/epfl/skysync/components/GroupChat.kt
@@ -39,10 +39,12 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.input.ImeAction
+import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import ch.epfl.skysync.models.message.GroupDetails
 import ch.epfl.skysync.models.message.MessageDateFormatter
+import ch.epfl.skysync.ui.theme.Purple40
 import ch.epfl.skysync.ui.theme.lightGray
 import ch.epfl.skysync.ui.theme.lightOrange
 
@@ -58,11 +60,12 @@ import ch.epfl.skysync.ui.theme.lightOrange
 fun GroupChat(
     groupList: List<GroupDetails>,
     onClick: (GroupDetails) -> Unit,
-    paddingValues: PaddingValues
+    paddingValues: PaddingValues,
+    adminBoolean: Boolean
 ) {
   var searchQuery by remember { mutableStateOf("") }
-  Column(modifier = Modifier.fillMaxSize().padding(paddingValues).padding(horizontal = 16.dp)) {
-    GroupChatTopBar()
+  Column(modifier = Modifier.fillMaxSize().padding(paddingValues).padding(16.dp)) {
+    GroupChatTopBar(adminBoolean, paddingValues)
     Spacer(modifier = Modifier.fillMaxHeight(0.02f))
 
     OutlinedTextField(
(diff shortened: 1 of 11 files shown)


app/src/main/java/ch/epfl/skysync/screens/admin/AdminFlightDetailsScreen.kt
```diff
@@ -150,7 +150,7 @@ fun FlightDetailBottom(
                   .testTag("ConfirmButton")) {
             TextButton(
                 onClick = { confirmClick() },
-                shape = RoundedCornerShape(topEnd = 8.dp, bottomEnd = 8.dp),
+                shape = RoundedCornerShape(topEnd = 16.dp, bottomEnd = 16.dp),
```
RAFInery, line 153: maybe that kind of parameters (rounded shape of button) could be unified in a style sheet (like we did with the colors)

app/src/main/java/ch/epfl/skysync/components/GroupChat.kt
```diff
@@ -58,11 +60,12 @@ import ch.epfl.skysync.ui.theme.lightOrange
 fun GroupChat(
     groupList: List<GroupDetails>,
     onClick: (GroupDetails) -> Unit,
-    paddingValues: PaddingValues
+    paddingValues: PaddingValues,
+    adminBoolean: Boolean
```
RAFInery, line 64: `isAdmin` instead of `adminBoolean` would be more explanatory

app/src/main/java/ch/epfl/skysync/screens/admin/UserManagement.kt
```diff
@@ -125,40 +128,53 @@ fun SearchBar(query: String, onQueryChanged: (String) -> Unit) {
...
+      DropdownMenu(expanded = expanded, onDismissRequest = { expanded = false }) {
         DropdownMenuItem(
             onClick = {
-              displayText = role.name
-              onRoleSelected(role)
+              displayText = "Filter by role"
+              onRoleSelected(null)
               expanded = false
             },
-            text = { Text(role.name) })
```
RAFInery, line 159: use `role.description `instead of` role.name` to have a nicer string representation to display

app/src/main/java/ch/epfl/skysync/screens/admin/UserManagement.kt
```diff
@@ -125,40 +128,53 @@ fun SearchBar(query: String, onQueryChanged: (String) -> Unit) {
...
+            text = { Text("All Roles") })
+        roles.forEach { role ->
+          DropdownMenuItem(
+              onClick = {
+                displayText = role.name
```
RAFInery, line 162: `role.description`

app/src/main/java/ch/epfl/skysync/screens/admin/UserManagement.kt
```diff
@@ -125,40 +128,53 @@ fun SearchBar(query: String, onQueryChanged: (String) -> Unit) {
...
+                onRoleSelected(role)
+                expanded = false
+              },
+              text = { Text(role.name) })
```
RAFInery, line 166: `role.description`

app/src/main/java/ch/epfl/skysync/screens/crewpilot/Chat.kt
```diff
@@ -34,7 +34,8 @@ fun ChatScreen(navController: NavHostController, viewModel: ChatViewModel) {
             onClick = { selectedGroup ->
               navController.navigate(Route.CREW_TEXT + "/${selectedGroup.id}")
             },
-            paddingValues = padding)
+            paddingValues = padding,
+            false)
```
RAFInery, line 38: always more readable to put the parameter name i.e 
`adminBoolean = false` or alternatively define `val isAdmin = false `and use `isAdmin` instead of just `false`

RAFInery (approved):
added some small comments but the rest looks great

app/src/main/java/ch/epfl/skysync/screens/UserDetails.kt
```diff
@@ -35,10 +55,43 @@ fun UserDetailsScreen(navController: NavHostController, flightsViewModel: Flight
...
-      FlightsList(userFlights.value, lightOrange, paddingValues, "Completed Flights") {
-          selectedFlight ->
+      FlightsList(userFlights.value, paddingValues) { selectedFlight ->
         navController.navigate(Route.ADMIN_FLIGHT_DETAILS + "/$selectedFlight")
       }
     }
   }
 }
+
+@Composable
```
mathieumognol, line 65: Why did you redefine `FlightList` here? Because there is already a component `FlightList` that does the same thing I think

mathieumognol (approved):
LGTM

app/src/main/java/ch/epfl/skysync/screens/UserDetails.kt
Zep3k: Because FlightList comes with a colored header and i think it look weird here so i redefined it to just have the flights

Zep3k: now i create function in FlightList called FlightListContent that only has the list and i call iton UserDetails instead of redefining FlightList

PR #397, opened 2024-05-23
Filter finished flights

RAFInery:
# filter `FinishedFlight` for `COMPLETED` and `MISSING_REPORT`
- added function to update the `FlightStatus` of a `FinishedFlight` as a function of a user
For an `Admin` a `FinishedFlight` is considered `COMPLETED` once all team members have submitted their `Report`
For a `Crew/Pilot `it requires only the submission of the personal `Report` to have a `COMPLETED` finished flight
-

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
+++ b/app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
@@ -13,6 +13,7 @@ import ch.epfl.skysync.models.calendar.TimeSlot
 import ch.epfl.skysync.models.flight.Balloon
 import ch.epfl.skysync.models.flight.Basket
 import ch.epfl.skysync.models.flight.ConfirmedFlight
+import ch.epfl.skysync.models.flight.FinishedFlight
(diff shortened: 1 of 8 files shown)


app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
```diff
@@ -106,9 +107,16 @@ class FlightsViewModel(
           Log.d("FlightsViewModel", "Admin user loaded")
           _currentFlights.value = repository.flightTable.getAll(onError = { onError(it) })
         } else if (_currentUser.value is Pilot || _currentUser.value is Crew) {
-          _currentFlights.value =
+          val fetchedFlights =
               repository.userTable.retrieveAssignedFlights(
                   repository.flightTable, userId ?: UNSET_ID, onError = { onError(it) })
```
Plouc314, line 112: It was missed during a previous review, but `userId` should not be nullable (take `userId: String` as attribute), thus there will be no need for `userId ?: UNSET_ID` which wouldn't work anyway.

Zep3k (approved):
LGTM

PR #418, opened 2024-05-24
Fix available equipement

### LINKED ISSUE #340: Fix flight viewmodel
* When modifying the date of a flight the selected user should not be selected if not available on this new date
* Filter available user according to their role

Prolas:
## Bug
When modifying a flight, one could select a new date, and confirm even if the team, balloon or basket are not available.

## Solutions
On date or timeslot changes, check if the user/equipment are available. If not display an error on the field.

## UX enhancement
Addition of deletion button for balloon and basket to unset a balloon or basket.

close #340


app/src/main/java/ch/epfl/skysync/screens/admin/ModifyFlight.kt
```diff
@@ -16,18 +16,25 @@ fun ModifyFlightScreen(
...
+  val allBalloons = viewModel.currentBalloons.collectAsStateWithLifecycle()
+  val allBaskets = viewModel.currentBaskets.collectAsStateWithLifecycle()
+  val allVehicles = viewModel.currentVehicles.collectAsStateWithLifecycle()
+  val availableUsers = viewModel.availableUsers.collectAsStateWithLifecycle()
   val allRoleTypes = RoleType.entries
+
+  if (flightToModify.value != null) {
+    viewModel.setFlight(flightToModify.value!!)
+    println(flightToModify)
+    println(allBaskets)
```
Plouc314, line 30: Leftover prints

app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
```diff
@@ -133,12 +139,18 @@ class FlightsViewModel(
                   localDate = date!!,
                   timeslot = timeSlot!!,
                   onError = { onError(it) })
+
+          if (_flight.value?.balloon != null && firstFill) {
```
Plouc314, line 143: Instead of checking if it is the first time it is refreshed, you should check if the date is the date of the flight, that way you can add the flight members as available users when the date match and not otherwise

app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
```diff
@@ -133,12 +139,18 @@ class FlightsViewModel(
...
+            val availableBalloons = _availableBalloons.value.filter { it != _flight.value?.balloon }
+            _availableBalloons.value = availableBalloons.plus(_flight.value?.balloon!!)
+            println(_availableBalloons.value)
```
Plouc314, line 146: Leftover print

app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
```diff
@@ -147,26 +159,38 @@ class FlightsViewModel(
...
+          }
         } else {
           _availableUsers.value = repository.userTable.getAll(onError = { onError(it) })
         }
       }
 
-  fun refreshAvailableVehicles() =
+  fun refreshAvailableVehicles(firstFill: Boolean = false) =
       viewModelScope.launch {
+        println("REFRESH VEHICLES")
```
Plouc314, line 174: Leftover print but in upper case

Prolas: I will change it

app/src/main/java/ch/epfl/skysync/viewmodel/FlightsViewModel.kt
Prolas: Good catch. I will change it

Plouc314 (approved):
LGTM

PR #430, opened 2024-05-25
Fix display flight bug

### LINKED ISSUE #419: Fix display flight trace
The `confirmedFlights` in the `InFlightViewModel` are only the ones of today, which doesn't work for past flight

Plouc314:
* Fixed a bug that would only find the flight whose flight trace to display if it had taken place today
* Found a bug, when clicking on the display flight trace button as admin, we navigate on the flight screen, which is on the crew/pilot graph, so we end up on the crew/pilot graph

Close #419


Prolas (approved):
LGTM

PR #432, opened 2024-05-26
Refactor button

Baribal02:
The button is now at the bottom of the screen
There is also a confirm pop up when the pilot wants to stop the flight
image

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/MainActivity.kt
+++ b/app/src/main/java/ch/epfl/skysync/MainActivity.kt
@@ -22,14 +22,14 @@ import com.google.firebase.auth.FirebaseAuth
 class MainActivity : ComponentActivity() {
   private lateinit var signInLauncher: ActivityResultLauncher<Intent>
   private var userGlobalViewModel: UserGlobalViewModel? = null
-  private val db: FirestoreDatabase = FirestoreDatabase(true)
+  private val db: FirestoreDatabase = FirestoreDatabase()
   private val repository: Repository = Repository(db)
 
   private fun onSignInResult(result: FirebaseAuthUIAuthenticationResult) {
     val vm = userGlobalViewModel ?: return
     if (result.resultCode == RESULT_OK) {
       val user = FirebaseAuth.getInstance().currentUser!!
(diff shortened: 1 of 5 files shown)


app/src/main/java/ch/epfl/skysync/MainActivity.kt
```diff
@@ -22,14 +22,14 @@ import com.google.firebase.auth.FirebaseAuth
...
   private var userGlobalViewModel: UserGlobalViewModel? = null
-  private val db: FirestoreDatabase = FirestoreDatabase()
+  private val db: FirestoreDatabase = FirestoreDatabase(true)
   private val repository: Repository = Repository(db)
 
   private fun onSignInResult(result: FirebaseAuthUIAuthenticationResult) {
     val vm = userGlobalViewModel ?: return
     if (result.resultCode == RESULT_OK) {
       val user = FirebaseAuth.getInstance().currentUser!!
-      vm.loadUser(user.uid, user.email!!)
```
Plouc314, line 32: Looks like leftovers from debugging

app/src/main/java/ch/epfl/skysync/screens/crewpilot/Flight.kt
```diff
@@ -228,6 +219,36 @@ fun FlightScreen(
                           .background(color = Color.White, shape = RoundedCornerShape(8.dp))
                           .padding(6.dp))
             }
+            var showConfirmationDialog by remember { mutableStateOf(false) }
```
Plouc314, line 222: This should be moved up with the other variables

Plouc314 (approved):
LGTM

app/src/main/java/ch/epfl/skysync/MainActivity.kt
Baribal02: You're right. It's now removed

app/src/main/java/ch/epfl/skysync/screens/crewpilot/Flight.kt
Baribal02: Moved it !

PR #446, opened 2024-05-27
Delete group chat

### LINKED ISSUE #407: Admin can delete chatGroup

Zep3k:
added a button that appears after you hold a group to delete that group
Closes #407 
Screenshot_20240527-160419_SkySync


Plouc314:
You should update `ChatViewModel.messageGroups` to remove from it the group that has been removed (that way you don't have to refresh the message groups), also at the moment one could send a message to a group that has been deleted, but this is not dramatic (it is just an unreachable entry in the message table)

hacknstack (approved):
LGTM

PR #448, opened 2024-05-27
assigned flight calendar fix

### LINKED ISSUE #428: Availability on assigned flights may not be changed.
When a user is assigned to a flight he should not be able to modify the availability

hacknstack:
closes #428 
basically a rework of #342 in the new VM environement


app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
```diff
@@ -79,12 +81,22 @@ class CalendarViewModel(
    */
   private suspend fun refreshUserAndCalendars() {
     _user.value = userTable.get(uid, this::onError)!!
-    val currentAvCal = AvailabilityCalendar(userTable.retrieveAvailabilities(uid, this::onError))
+    var currentAvCal = AvailabilityCalendar(userTable.retrieveAvailabilities(uid, this::onError))
+    val flights = userTable.retrieveAssignedFlights(flightTable, uid, this::onError)
+    flights.forEach { flight ->
+      if (flight !is FinishedFlight) {
```
RAFInery, line 87: status should also be set to `assigned` for any flight when affected not just for a `FinishedFlight`

RAFInery (changes requested):
It would be nicer to update and persist the assigned status of the availability upon affection.  Like this it does not have to be checked on every load.

app/src/main/java/ch/epfl/skysync/viewmodel/CalendarViewModel.kt
hacknstack: It's set to assigned for every flight that isn't finished

RAFInery (approved):
LGTM

PR #451, opened 2024-05-29
fixed location bug

hacknstack:
When click on the flight takeoff and landing location, it now takes you to the adequate coordinates.


RAFInery (approved):
LGTM

PR #453, opened 2024-05-29
Report fix

### LINKED ISSUE #429: Fix FinishFlightDetails logic to have add report

hacknstack:
closes #429,
Since we still have the Admin graph bug in stats screen, the see report shows the report of everyone always.

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/screens/crewpilot/FlightDetailsScreen.kt
+++ b/app/src/main/java/ch/epfl/skysync/screens/crewpilot/FlightDetailsScreen.kt
@@ -37,23 +37,22 @@ fun CrewFlightDetailScreen(
           is FinishedFlight -> {
             FinishedFlightDetailBottom(
                 reportClick = {
-                    val reportDone = (flight as FinishedFlight).reportId.any { report ->
+                  val reportDone =
+                      (flight as FinishedFlight).reportId.any { report ->
                         report.author == viewModel.userId
+                      }
+                  if (reportDone) {
+                    navController.navigate(Route.REPORT + "/${flightId}")
+                  } else {
+                    if ((flight as FinishedFlight)
+                        .team
+                        .hasUserRole(RoleType.PILOT, viewModel.userId!!)) {
+                      navController.navigate(Route.PILOT_REPORT + "/${flightId}")
+                    } else {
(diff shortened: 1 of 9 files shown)


app/src/main/java/ch/epfl/skysync/screens/crewpilot/FlightDetailsScreen.kt
```diff
@@ -35,7 +36,23 @@ fun CrewFlightDetailScreen(
           }
           is FinishedFlight -> {
             FinishedFlightDetailBottom(
-                reportClick = { navController.navigate(Route.REPORT + "/${flightId}") },
+                reportClick = {
+                  val reportDone =
```
RAFInery, line 40: would be cleaner to have the whole logic in the ViewModel. Here you could add a function to the VM like for example:
`reportDoneFor(flightid, userId): Boolean` that can then simply be accessed through the VM

app/src/main/java/ch/epfl/skysync/screens/reports/ReportDetailsScreen.kt
```diff
@@ -37,8 +38,10 @@ fun ReportDetailsScreen(
     val display = (isAdmin && reportIds!!.isNotEmpty()) || reportIds!!.any { (it.author == userId) }
     Column() {
       CustomTopAppBar(navController, "Report")
+      reportIds!!.forEach { report -> Log.d("report", report.author) }
       if (display) {
-        ReportDetail(users, reportIds, isAdmin, userId)
+        val reportLists = if (isAdmin) reportIds else reportIds!!.filter { (it.author == userId) }
```
RAFInery, line 43: same as above. Ideally, the VM (which know the user) will return the relevant reportIds

RAFInery (approved):
I let few comments but apart from that looks fine

app/src/main/java/ch/epfl/skysync/screens/reports/ReportDetailsScreen.kt
hacknstack: Big iq comment ty

app/src/main/java/ch/epfl/skysync/screens/crewpilot/FlightDetailsScreen.kt
hacknstack: 👍

PR #459, opened 2024-05-30
Add location input

Baribal02:
We can now change hte location for the landing and takeoff point
image

### DIFF
--- a/app/src/main/java/ch/epfl/skysync/components/forms/SpecialFields.kt
+++ b/app/src/main/java/ch/epfl/skysync/components/forms/SpecialFields.kt
@@ -150,6 +150,7 @@ fun LocationPickerField(
       onElementClick = {
         setLocation(it)
         query = if (it.name == "") it.latlng().toString() else it.name
+        active = false
       },
       propositions = result,
       showProposition = { if (it.name == "") it.latlng().toString() else it.name })
--- a/app/src/main/java/ch/epfl/skysync/components/forms/SpecialFields.kt
+++ b/app/src/main/java/ch/epfl/skysync/components/forms/SpecialFields.kt
@@ -30,6 +30,7 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import ch.epfl.skysync.components.GenericTimePicker
 import ch.epfl.skysync.components.RangeTimePicker
 import ch.epfl.skysync.components.SmallTitle
@@ -37,6 +38,7 @@ import ch.epfl.skysync.database.DateUtility.dateToHourMinuteString
 import ch.epfl.skysync.models.flight.Vehicle
 import ch.epfl.skysync.models.location.LocationPoint
 import ch.epfl.skysync.util.hasError
+import ch.epfl.skysync.viewmodel.FinishedFlightsViewModel
 import java.util.Date
 
 /**
@@ -125,11 +127,12 @@ fun LocationPickerField(
     location: LocationPoint,
     defaultPadding: Dp,
     title: String,
-    setLocation: (LocationPoint) -> Unit
+    viewModel: FinishedFlightsViewModel,
+    setLocation: (LocationPoint) -> Unit,
 ) {
-  // TODO implement search bar logic
   var active by remember { mutableStateOf(false) }
   var query by remember { mutableStateOf("") }
+  val result by viewModel.searchResults.collectAsStateWithLifecycle()
   Text(
       modifier = Modifier.fillMaxWidth().padding(horizontal = defaultPadding),
       text = title,
@@ -137,7 +140,10 @@ fun LocationPickerField(
   SearchBarCustom(
       title = title,
       query = query,
-      onQueryChange = { query = it },
+      onQueryChange = {
+        query = it
+        viewModel.getSearchLocation(it, location.time)
+      },
       onSearch = { active = false },
       active = active,
       onActiveChange = { active = it },
@@ -145,7 +151,7 @@ fun LocationPickerField(
         setLocation(it)
         query = if (it.name == "") it.latlng().toString() else it.name
       },
-      propositions = emptyList<LocationPoint>(),
+      propositions = result,
       showProposition = { if (it.name == "") it.latlng().toString() else it.name })
 }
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -115,6 +126,45 @@ class FinishedFlightsViewModel(val repository: Repository, val userId: String) :
...
+              responseBody
+            } catch (e: Exception) {
+              e.printStackTrace()
+              "error"
+            }
+      }
+
+  fun getSearchLocation(query: String, time: Int) =
+      viewModelScope.launch {
+        searchLocation(query).join()
```
Plouc314, line 147: You might want to make the `searchLocation` a suspend function, as you are already starting a coroutine in `getSearchLocation` and you immediately join it anyway

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -115,6 +126,45 @@ class FinishedFlightsViewModel(val repository: Repository, val userId: String) :
...
+        val gson = Gson()
+        if (_searchResponse.value == null) {
```
Plouc314, line 149: Does `_searchResponse` need to be a flow ? Couldn't you return the response from the `searchLocation` function, `_searchResults` is a flow and gets updated (also `_searchResponse` is private and never exposed)

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -115,6 +126,45 @@ class FinishedFlightsViewModel(val repository: Repository, val userId: String) :
...
+        } else if (_searchResponse.value != null && _searchResponse.value != "") {
+          Log.d("searchLocation", _searchResponse.value!!)
+          val jsonArray = gson.fromJson(_searchResponse.value, JsonArray::class.java)
+          Log.d("searchLocation", jsonArray.toString())
+          _searchResults.value =
+              jsonArray.map { jsonString ->
+                val jsonObject = gson.fromJson(jsonString, JsonObject::class.java)
+                val displayName = jsonObject.get("display_name")?.asString ?: ""
+                val latitude = jsonObject.get("lat")?.asDouble ?: 0.0
+                val longitude = jsonObject.get("lon")?.asDouble ?: 0.0
```
Plouc314, line 160: I'm not familiar with the `gson` library, can this throw an exception ? If the format is not the one expected we might want to display an error message instead of setting default values

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -115,6 +126,45 @@ class FinishedFlightsViewModel(val repository: Repository, val userId: String) :
...
+        val url = "https://nominatim.openstreetmap.org/search?q=$query&format=json&limit=4"
+        val request = Request.Builder().url(url).build()
+        _searchResponse.value =
+            try {
+              val response = client.newCall(request).execute()
+              val responseBody = response.body()?.string()
+              responseBody
+            } catch (e: Exception) {
+              e.printStackTrace()
+              "error"
```
Plouc314, line 141: Here we could call an `onError` method as in the other view models

app/src/main/java/ch/epfl/skysync/components/forms/SpecialFields.kt
```diff
@@ -125,27 +127,32 @@ fun LocationPickerField(
...
     defaultPadding: Dp,
     title: String,
-    setLocation: (LocationPoint) -> Unit
+    viewModel: FinishedFlightsViewModel,
+    setLocation: (LocationPoint) -> Unit,
 ) {
-  // TODO implement search bar logic
   var active by remember { mutableStateOf(false) }
   var query by remember { mutableStateOf("") }
+  val result by viewModel.searchResults.collectAsStateWithLifecycle()
```
Plouc314, line 135: Nitpick: `result` is a list so you can add an 's' `results`

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
```diff
@@ -115,6 +126,45 @@ class FinishedFlightsViewModel(val repository: Repository, val userId: String) :
...
+  fun getSearchLocation(query: String, time: Int) =
+      viewModelScope.launch {
+        searchLocation(query).join()
+        val gson = Gson()
+        if (_searchResponse.value == null) {
+          _searchResults.value = null
+        } else if (_searchResponse.value != null && _searchResponse.value != "") {
+          Log.d("searchLocation", _searchResponse.value!!)
+          val jsonArray = gson.fromJson(_searchResponse.value, JsonArray::class.java)
+          Log.d("searchLocation", jsonArray.toString())
```
Plouc314, line 154: Looks like debugging leftovers, you can remove them

Plouc314 (approved):
Great idea to use openstreetmap

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
Baribal02: Indead, it is now removed

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
Baribal02: Done

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
Baribal02: No, it is not needed. I have change that.

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
Baribal02: Yes, you're right. It is now the case.

app/src/main/java/ch/epfl/skysync/viewmodel/FinishedFlightsViewModel.kt
Baribal02: As openStreetMap always send a JsonArray as a string in case of a right response, there should be no error there. All the possible errors would happen during searchLocation

app/src/main/java/ch/epfl/skysync/components/forms/SpecialFields.kt
Baribal02: Changed it

PR #464, opened 2024-05-31
Fix available equipement and add tests

Prolas:
* Fixed a bug that displayed all equipment and users as available when adding a flight
* Added tests

### DIFF
--- a/app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
+++ b/app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
@@ -16,6 +16,7 @@ import ch.epfl.skysync.models.flight.Team
 import ch.epfl.skysync.models.location.FlightTrace
 import ch.epfl.skysync.models.location.LocationPoint
 import ch.epfl.skysync.models.reports.FlightReport
+import ch.epfl.skysync.models.user.User
 import java.time.LocalDate
 import java.time.LocalTime
 import kotlinx.coroutines.test.runTest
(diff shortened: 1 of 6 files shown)


app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
```diff
@@ -154,6 +155,108 @@ class FlightsViewModelTest {
...
+      assertEquals(expectedBaskets, currentBaskets.value.sortedBy { it.id })
+      assertEquals(expectedVehicles, currentVehicles.value.sortedBy { it.id })
+      assertEquals(expectedUsers, currentUsers.value.sortedBy { it.id })
+
+      viewModelAdmin.setDateAndTimeSlot(dbSetup.date1, TimeSlot.AM)
+
+      viewModelAdmin.refreshAvailableBalloons().join()
+      viewModelAdmin.refreshAvailableBaskets().join()
+      viewModelAdmin.refreshAvailableVehicles().join()
+      viewModelAdmin.refreshAvailableUsers().join()
```
Baribal02, line 195: The refresh are not necessary as it is already done in setDateAndTimeSlot but you need to wait that these coroutines are finished (which is what your code does in a way).

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
```diff
@@ -154,6 +155,108 @@ class FlightsViewModelTest {
...
+      assertEquals(expectedBaskets, currentBaskets.value.sortedBy { it.id })
+      assertEquals(expectedVehicles, currentVehicles.value.sortedBy { it.id })
+      assertEquals(expectedUsers, currentUsers.value.sortedBy { it.id })
+
+      viewModelAdmin.setDateAndTimeSlot(dbSetup.date2, dbSetup.date2TimeSlotInverse)
+
+      viewModelAdmin.refreshAvailableBalloons().join()
+      viewModelAdmin.refreshAvailableBaskets().join()
+      viewModelAdmin.refreshAvailableVehicles().join()
+      viewModelAdmin.refreshAvailableUsers().join()
```
Baribal02, line 243: Same as before

Baribal02 (approved):
LGTM

app/src/androidTest/java/ch/epfl/skysync/viewmodel/FlightsViewModelTest.kt
Prolas: Yes I know. Do you have a better solution ?

